PR #7, opened 2021-03-29
Proposal: remove a bunch of debugging features

kmod:
This PR removes
- tracemalloc
- _Py_CheckFunctionResult
- PyErr_BadInternalCall (and thus the checks surrounding it)
- recursion checking

Here are the perf results:
```
4086eaa3 26.66+-0.04              net -15.7% Add a --disable-debug-checks configure flag
0ed12549 26.52+-0.03  -0.5%+-0.2% net -16.2% Remove tracemalloc
7a1680e5 26.39+-0.04  -0.5%+-0.2% net -16.6% Remove _Py_CheckFunctionResult
0f35dd42 26.36+-0.02  -0.1%+-0.1% net -16.7% Remove PyErr_BadInternalCall
4fe43a54 26.29+-0.04  -0.3%+-0.1% net -16.9% Disable recursion checking
```

The numbers per change are somewhat variable (usually I see that tracemalloc is smaller and recursion checking is bigger) but the >1% total improvement seems consistent.

This is probably one of the biggest semantic changes that we'd be making so I think we should talk about it first.  I'm not really sure though how to get feedback about this, other than just putting out a release without them and seeing if people ask for them back.

### DIFF
--- a/Include/ceval.h
+++ b/Include/ceval.h
@@ -86,6 +86,10 @@ PyAPI_FUNC(int) Py_MakePendingCalls(void);
 PyAPI_FUNC(void) Py_SetRecursionLimit(int);
 PyAPI_FUNC(int) Py_GetRecursionLimit(void);
 
+#if !PY_DEBUGGING_CHECKS
+#define Py_EnterRecursiveCall(where) (0)
+#define Py_LeaveRecursiveCall(where) ((void)0)
+#else
 #define Py_EnterRecursiveCall(where)  \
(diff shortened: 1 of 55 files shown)


undingen (commented):
LGTM - this shows up as an incredible 2% speedup on my ryzen!
test_django is failing for me because of a difference in skipped tests
```
-FAILED (errors=1, skipped=917, expected failures=4)
+FAILED (errors=1, skipped=919, expected failures=4)
```

PR #11, opened 2021-04-26
add pyramid benchmark

undingen:
It's just the hello world example. But it seems to run about 50% faster (1104r/s vs 1644r/s)


undingen:
We have now a better pyramid test: kinto_bench closing this.

PR #20, opened 2021-05-08
Fix build errors on MSVC

ghost:
This wraps some gnu specific things so that people can build on MSVC. I wanted to test this interpreter out, but had to make these changes first.

### DIFF
--- a/Include/ceval.h
+++ b/Include/ceval.h
@@ -87,8 +87,8 @@ PyAPI_FUNC(void) Py_SetRecursionLimit(int);
 PyAPI_FUNC(int) Py_GetRecursionLimit(void);
 
 #if !PY_DEBUGGING_CHECKS
-#define Py_EnterRecursiveCall(where) (0)
-#define Py_LeaveRecursiveCall(where) ((void)0)
+#define Py_EnterRecursiveCall(...) (0)
+#define Py_LeaveRecursiveCall(..) ((void)0)
 #else
 #define Py_EnterRecursiveCall(where)  \
             (_Py_MakeRecCheck(PyThreadState_GET()->recursion_depth) &&  \
--- a/Include/ceval.h
+++ b/Include/ceval.h
@@ -88,7 +88,7 @@ PyAPI_FUNC(int) Py_GetRecursionLimit(void);
 
 #if !PY_DEBUGGING_CHECKS
 #define Py_EnterRecursiveCall(...) (0)
-#define Py_LeaveRecursiveCall(..) ((void)0)
+#define Py_LeaveRecursiveCall(...) ((void)0)
 #else
 #define Py_EnterRecursiveCall(where)  \
             (_Py_MakeRecCheck(PyThreadState_GET()->recursion_depth) &&  \
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -145,7 +145,16 @@ _PyObject_ArenaVirtualFree(void *ctx, void *ptr, size_t size)
 {
     VirtualFree(ptr, 0, MEM_RELEASE);
 }
+static void *
+_PyObject_ArenaMmap(void *ctx, size_t size){
+    return _PyObject_ArenaVirtualAlloc(ctx, size);
+}
 
+static void
+_PyObject_ArenaMunmap(void *ctx, void *ptr, size_t size)
+{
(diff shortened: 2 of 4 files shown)


Python/ceval.c
```diff
@@ -5012,8 +5024,12 @@ call_function_ceval(PyThreadState *tstate, PyObject ***pp_stack, Py_ssize_t opar
 #else
     PyObject **stack = (*pp_stack) - nargs - nkwargs;
 #endif
-
+#ifndef _MSC_VER
     if (__builtin_expect(tstate->use_tracing, 0)) {
+#else
+     if ((tstate->use_tracing, 0)) {
```
kmod: I think you need to remove the `0` (and parenthesis) here. Also the whitespace seems a little bit funky

Include/ceval.h
```diff
@@ -87,8 +87,8 @@ PyAPI_FUNC(void) Py_SetRecursionLimit(int);
 PyAPI_FUNC(int) Py_GetRecursionLimit(void);
 
 #if !PY_DEBUGGING_CHECKS
-#define Py_EnterRecursiveCall(where) (0)
-#define Py_LeaveRecursiveCall(where) ((void)0)
+#define Py_EnterRecursiveCall(...) (0)
+#define Py_LeaveRecursiveCall(..) ((void)0)
```
kmod: I'm curious, why is this needed? Also are there supposed to be 3 dots and then 2 dots?

Objects/obmalloc.c
```diff
@@ -145,7 +145,16 @@ _PyObject_ArenaVirtualFree(void *ctx, void *ptr, size_t size)
 {
     VirtualFree(ptr, 0, MEM_RELEASE);
 }
+static void *
```
kmod, line 148: I see what you're doing here but I think a better fix might be to change the callers of _PyObject_ArenaMmap to call the correct function

Python/sysmodule.c
```diff
@@ -1732,9 +1732,13 @@ static Py_ssize_t
 sys_getallocatedblocks_impl(PyObject *module)
 /*[clinic end generated code: output=f0c4e873f0b6dcf7 input=dab13ee346a0673e]*/
 {
+    #if PY_DEBUGGING_FEATURES
```
kmod, line 1735: I'm curious, what's the link error here?

kmod (changes requested):
Thanks so much for looking into this! I left some comments on your changes.

I think the bigger comment is that, sadly, windows support will be much harder than just fixing the build errors, since our JIT will need to be updated and I'm not sure what else. But this seems like a good first step!

Include/ceval.h
ghost: Errors out on wrong number of arguments to the macro on MSVC.
The two dots is my mistake

Python/sysmodule.c
ghost: sys_getallocatedblocks_impl gets referenced later despite being ifdefed out. I stubbed to remove the error instead of removing the reference since its a documented member of sys

Objects/obmalloc.c
ghost: I think just having a single wrapper around virtual memory allocation/deallocation is more convenient and makes more sense overall.

ghost:
I moved over just the changes I had that were related to getting it to compile to this repo, I seem to have made two mistakes doing so, those are now corrected.

Include/ceval.h
kmod: What I meant was "why is a difference from CPython needed here", and it looks like the answer is because I messed up the conversion of Py_LeaveRecursiveCall. I think the proper solution here is to keep the definition of Py_EnterRecursiveCall and to remove the erroneous argument from Py_LeaveRecursiveCall

Objects/obmalloc.c
kmod: Sure it could be a function wrapper or a macro wrapper, but I think it's incorrect to say that Mmap is implemented using VirtualAlloc.

Python/sysmodule.c
kmod: I'm confused about what's going on here, is MSVC trying to compile ifdef'd out code? What's the error you get?

I removed the function because I believe it's better to not have the function than to have it behave incorrectly. It looks like the removal was unnecessary, so if you want you could just remove the `PY_DEBUGGING_FEATURES` ifdefs.

PR #28, opened 2021-05-11
kinto (uwsgi+pyramid) benchmark

kmod:
kinto is a project built on top of pyramid
run it using nginx+uwsgi


undingen (approved):
LGTM
I was waiting for the embedding support but could already test it with the WIP patches,

PR #44, opened 2021-06-07
update bolt

undingen:
While tracking down the instrumentation crash on the shared object I noticed
that bolt switched development to a branch called 'rebased' which uses a llvm monorepo.
The current master branch we used before seems not to be maintained anymore.

I did not measure any performance difference.


kmod:
Ooh awesome, maybe we could have just a single version of llvm now

PR #45, opened 2021-06-08
Remove PyType_Ready mixed allocated subclass error

undingen:
We disable this error in pyston because cython has to workaround it
by setting temporarily `Py_TPFLAGS_HEAPTYPE` on all `PyTypeObjects` which
causes further issues for pyston.
So instead we remove the error and don't pass the flag in cython.

https://github.com/cython/cython/issues/4200


kmod:
:+1:

PR #46, opened 2021-06-09
Build libpython.a and libpython.so

kmod:
- Makefile cleanup
- Rename everything from "pyston3.8" to "python3.8-pyston2.2"
  - This better supports things that expect the "python$(VERSION)" format, as well as prevents conflicts between multiple versions of pyston targeting the same version of Python
- Build a separate PIC version of aot
- split opt into "opt" (runnable from your dev environment) and "release" (only works from /usr)
- support --enable-shared builds of cpython
- Mix the shared and static builds in the same way that ubuntu does (copy libpython.so and _sysconfigdata.py on top of the static build)

I tested that this makes uwsgi work and is decently faster

waiting on:
- `make package` test
- performance test

Todo:
- get bolt working on the shared object
- test the full release process


undingen (commented):
:fireworks: looking very good! Thanks - finally embedding support while also cleaning up the makefile a bit :)

Everything I tried except debian package build seemed to work fine.
I did add https://github.com/pyston/pyston/pull/47 which fixes a recent problem I introduced which brings it further but unfortunately the bolt instrumented executable is crashing inside the docker container?!? 
I can investigate tomorrow why it's crashing if you want.

kmod:
Are you getting a consistent crash? I did see a segfault once but I haven't been able to reproduce it, I'll look into it but if you have a consistent crash I could use some help

PR #47, opened 2021-06-10
fix dockerfiles after bolt update

undingen:
- 16.04 and 18.04 come with an to old cmake - use newest cmake directly from kitware
20.04 is new enough so I just used the stock ubuntu one (which means it will be older then the one on 16.04 and 18.04)

- remove pyston/bolt/llvm stuff


kmod:
:+1:

PR #48, opened 2021-06-10
fix: dont install uwsgi for the pgo task

kmod:
We're not ready to build uwsgi yet because we haven't built the shared build.

This was working on my machine because I had a cached uwsgi but when I went to a
benchmarking machine it started failing.


undingen (approved):
LGTM (was a good reminder that I have to test your patch with cleared pip cache :worried:)

PR #49, opened 2021-06-14
Backport -fno-semantic-interposition from cpython

undingen:
In https://bugs.python.org/issue38980 they mention that they get a 1.3x speedup with `-fno-semantic-interposition` for the `--enable-shared` build. Now that we support soon this build type it likely makes sense to use the same.
I did not measure that big speedup but nbody improves from 8.78s -> 8.59s on the optshared pyston binary.

Original commit:
bpo-38980: Add -fno-semantic-interposition when building with optimizations (GH-22862)


kmod:
good idea!

PR #50, opened 2021-06-15
fix release packaging

undingen:
- llvm-bolt crashes while instrumenting when build via `dpkg-buildpackage`. Building it before fixes the issue.
- make_portable_dir.py: change dir from `/usr/lib/pyston3.8/lib-dynload/` to `/usr/lib/python3.8-pyston2.2/lib-dynload/`

With this changes I can create release packages again.


kmod:
:+1:

what a weird issue with bolt

PR #51, opened 2021-06-17
JIT: support PIC compilation

undingen:
the JIT always emits code to address which fit into 32bit
but if PIC is enabled non JIT code may use a larger address space.
This causes issues because x86_64 rip memory access only use 32bit offsets.
To solve this issue we have to load the pointer into a register.


kmod:
awesome, thanks!

PR #52, opened 2021-06-19
Smill fixes for libpython packaging

kmod:
Couple minor things I didn't get quite right about packaging up the libpython files


undingen (approved):
LGTM seems to work. 
I'm just not sure what the difference between `libpython3.8-pyston2.2.so` and `libpyston3.so` is? They same to not be the same and we seem to distribute both.
EDIT: a okay `libpyston3.so` seems to be a wrapper around the other one?

kmod:
I just copied what cpython does -- it looks like libpython3.so exists just to have a dependency on libpython3.8.so in case you don't care about the version of python. I'm not sure why they didn't do that with a symlink, though, which I thought was more traditional.

kmod:
Or rather, in a previous change I had changed cpython's behavior, and this PR puts it back

PR #53, opened 2021-06-21
Backport tp_vectorcall changes and improve AOT traces

undingen:
This example:
```python
def f():
    l = [1,2,3]
    for i in range(10000000):
        type(l)
        list(l)
        tuple(l)
        set(l)
        float(42)
        bool(1)
        range(1)
f()
```
Takes now:
```
cpython 3.8.5: 5.41s
pyston before: 2.54s
pyston new   : 1.94s
```

- generated code is good with the change to pystol
- added traces for `list`, `tuple` and `set`
- had to change most patches slightly e.g. `/* static */` or where they depend on previous unrelated patches.
- I did not backport all of them: e.g. `frozenset` and `dict` are missing
- our testsuite is currently failing on my machine (cpython testsuite is working) so I could not test it on everything yet
-  should I add `PYSTON_SPEEDUPS` ifdefs around all changes?

### DIFF
--- a/Include/modsupport.h
+++ b/Include/modsupport.h
@@ -71,9 +71,12 @@ PyAPI_FUNC(int) _PyArg_UnpackStack(
     ...);
 
 PyAPI_FUNC(int) _PyArg_NoKeywords(const char *funcname, PyObject *kwargs);
+PyAPI_FUNC(int) _PyArg_NoKwnames(const char *funcname, PyObject *kwnames);
 PyAPI_FUNC(int) _PyArg_NoPositional(const char *funcname, PyObject *args);
 #define _PyArg_NoKeywords(funcname, kwargs) \
     ((kwargs) == NULL || _PyArg_NoKeywords((funcname), (kwargs)))
+#define _PyArg_NoKwnames(funcname, kwnames) \
+    ((kwnames) == NULL || _PyArg_NoKwnames((funcname), (kwnames)))
 #define _PyArg_NoPositional(funcname, args) \
     ((args) == NULL || _PyArg_NoPositional((funcname), (args)))
 
--- a/Lib/test/test_float.py
+++ b/Lib/test/test_float.py
@@ -64,6 +64,9 @@ class GeneralFloatCases(unittest.TestCase):
         # See bpo-34087
         self.assertRaises(ValueError, float, '\u3053\u3093\u306b\u3061\u306f')
 
+    def test_noargs(self):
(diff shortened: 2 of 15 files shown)


kmod:
lgtm! though quick question -- why do we need to mark ob_base->ob_type as constant? Just felt a bit odd where that comes from, should we mark the whole base type as constant as well?

undingen:
I had to make the base type static so it can optimize away this access:
```C
static inline vectorcallfunc
_PyVectorcall_Function(PyObject *callable)
{
    PyTypeObject *tp = Py_TYPE(callable);
    Py_ssize_t offset = tp->tp_vectorcall_offset;
    vectorcallfunc ptr;
    if (!PyType_HasFeature(tp, _Py_TPFLAGS_HAVE_VECTORCALL)) {
        return NULL;
    }
    assert(PyCallable_Check(callable));
    assert(offset > 0);
    memcpy(&ptr, (char *) callable + offset, sizeof(ptr));
    return ptr;
}
```
E.g. for `float(42)`:
It calls `_PyVectorcall_Function(&PyFloat_Type)`
With `ob_base` and `tp_vectorcall_offset` marked constant it figures out that `Py_TYPE(callable)` is `&PyType_Type` and that `PyType_Type.tp_vectorcall_offset` is `offsetof(PyTypeObject, tp_vectorcall)`. It now reads the types `PyFloat_Type.tp_vectorcall` which is also const `float_vectorcall`...


------

It does look strange that accessing the field is `x->ob_base.ob_base.ob_type` not only `x->ob_base.ob_type` but this is because the types base type is declared as a `PyVarObject` and not a `PyObject`.

```C

typedef struct _typeobject {
    PyObject_VAR_HEAD
    ...
    
#define PyObject_VAR_HEAD      PyVarObject ob_base;

typedef struct {
    PyObject ob_base;
    Py_ssize_t ob_size; /* Number of items in variable part */
} PyVarObject;

typedef struct _object {
    _PyObject_HEAD_EXTRA
    Py_ssize_t ob_refcnt;
    struct _typeobject *ob_type;
} PyObject;
```

And I think we don't need to make the base whole type constant because I assume it's already done by `addConstTypeAndConstSubclasses()`

Hope this clears up things. And let me know if I have a thinko.

undingen:
I added a new commit which removes a unnecessary indirection for the CALL_* traces and fixes the `aot_trace_only` make target

kmod:
lgtm!

PR #56, opened 2021-06-28
Add unboxed indexing traces

kmod:
I added PyObject_GetItemLong, which is an optimized version of PyObject_GetItem but can only be called with a 64-bit long as its argument, and the unboxed value must be passed as well, and set up the jit to call this function if the index is seen to be a constant int.

By itself this isn't really any faster since it will just unbox the long and use that. So I added some optimization passes that replace the unboxing operation with a usage of the passed unboxed value. I also added an exception-check-removal pass that is rudimentary but simplifies the generated code.

I'm not happy that a number of assumptions are now encoded into the optimization passes, when I think they should be explicitly stated as assumptions by aot_gen. Unfortunately the version of clang we're using has a bug where if you do `__builtin_assume(foo())` it will leave a call to `foo()` in the generated assembly. So in a future PR I'll update clang and switch this to using __builtin_assume instead of hardcoded-assumptions.

I also sped up non-constant tuple[int] indexing.

### DIFF
--- a/Python/aot_ceval_jit.c
+++ b/Python/aot_ceval_jit.c
@@ -1061,7 +1061,9 @@ static void deferred_vs_emit(Jit* Dst) {
 // Look at the top value of the value stack and if its a constant return the constant value,
 // otherwise return NULL
 static PyObject* deferred_vs_peek_const(Jit* Dst) {
-    assert(Dst->deferred_vs_next > 0);
+    if (Dst->deferred_vs_next == 0)
+        return NULL;
+
     DeferredValueStackEntry* entry = &Dst->deferred_vs[Dst->deferred_vs_next - 1];
     if (entry->loc == CONST)
         return PyTuple_GET_ITEM(Dst->co_consts, entry->val);
--- a/Python/aot_ceval_jit.c
+++ b/Python/aot_ceval_jit.c
@@ -1058,6 +1058,16 @@ static void deferred_vs_emit(Jit* Dst) {
     }
 }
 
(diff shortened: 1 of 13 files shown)


Python/aot_ceval_jit.c
```diff
@@ -1058,6 +1058,16 @@ static void deferred_vs_emit(Jit* Dst) {
     }
 }
 
+// Look at the top value of the value stack and if its a constant return the constant value,
+// otherwise return NULL
+static PyObject* deferred_vs_peek_const(Jit* Dst) {
+    assert(Dst->deferred_vs_next > 0);
```
undingen: this assert could be triggered if for some rare cases the deferred value stack got cleared e.g. call is jump target, deferred value stack got too large etc. I suggest changing it to `if (Dst->deferred_vs_next == 0) return NULL;`

undingen (commented):
Looking good :+1:, I only have a small comment about the `assert` inside `deferred_vs_peek_const` and I'm not 100% sure if the assumptions of the `ExceptionTrackingPass` are correct but I guess it should be okay for our traced functions. Did you check which exception checks it's removing? Is it only triggering on the new `PyObject_GetItemLong*` traces or did it also find other cases?

kmod:
It only fired in two other cases: callFloat and callRange. float does something like
```c
if (arg0 == NULL) {
  if (!PyErr_Occurred())
    PyErr_SetString("null argument");
  return NULL;
}
```
and the pass set PyErr_Occurred to NULL.

Once you add your args-are-not-null optimization then it won't catch any :)

PR #59, opened 2021-06-29
AOT: add __builtin_assume() for CALL_{FUNCTION,METHOD} traces

undingen:
this eliminates a few NULL pointer checks.
E.g. `call_function_ceval_no_kwFloat2` will remove one (verified by looking at the IR code and disassembly)

This was easier then I expected :smile:


kmod:
awesome!

PR #60, opened 2021-06-30
Pystol: remove refcount changes to immortal objects

undingen:
If we know that always the same immortal object is decref/increfed we can
remove the update.
E.g for `call_function_ceval_no_kwisinstanceLong3` we have a `Py_DECREF(stack[-oparg+1])`
but we do guard that `stack[-oparg+1] == (PyObject*)&PyLong_Type` so we know that we can remove it.

This is triggering on a lot of our traces.
It's currently not yet triggering for `builtin_isinstance_obj` because the python object
is heap allocated but I'm trying to solve this in the near future.

### DIFF
--- a/Objects/object.c
+++ b/Objects/object.c
@@ -1838,12 +1838,22 @@ PyObject _Py_NotImplementedStruct = {
 PyStatus
 _PyTypes_Init(void)
 {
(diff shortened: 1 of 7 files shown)


kmod:
Awesome!

PR #61, opened 2021-07-02
AOT: trace into builtin isinstance, len, ord 

undingen:
Unfortunately even after many many hours spend I was not able to expand nitrous to automatically
constant fold the constant builtin PyFunctionObjects. I tried different ways: facts, custom pass... but nothing worked.

The only robust way I found which optimizes everything away is by not heap allocating the PyCFunctionObjects.
So this is how I implemented it now. 
It should be easy to add more builtins.

- isinstance: before we stopped at 'builtin_isinstance' we now trace it completely
- isinstance: give better tracing examples and verify they return what expected
- len: added a few __builtin_assume/__builtin_unreachable checks which removes so unnecessary error handling


kmod:
awesome!

PR #66, opened 2021-07-08
add PYSTON_VERSION_HEX

undingen:
Similar to `PY_VERSION_HEX` but for the pyston version number, should make it easier for 3rd party code to check for a specific version number.
I was worried that we run into issues with exisiting software using it to check for pyston_v1 (e.g. old cythenized code) but it looks like we defined `PYSTON_VERSION` not `PYSTON_VERSION_HEX` so should be fine (github code search found no code in any repo checking it).


kmod:
:+1:

PR #67, opened 2021-07-08
Apply bolt to our .so files

kmod:
Tested that this is about 5.5% faster for kinto_bench_unopt


undingen (approved):
Works good on my intel laptop :+1:.
How much work would it to be an option (Makefile.local?) to disable the perf bolt run until we can use .so instrumentation? It would be nice to still be able to built a opt build on my amd desktop.

undingen:
ok mmh it seems unopt build is also using bolt? I'm getting this error:
```
LD_LIBRARY_PATH=:/home/marius/dev/pyston_v2/pyston/build/cpython_unoptshared_install/usr/lib LD_PRELOAD=libpython3.8-pyston2.2.so.1.0.prebolt perf record -e cycles:u -j any,u -o pyston/build/cpython_unoptshared_install/usr/lib/libpython3.8-pyston2.2.so.1.0.perf -- /tmp/tmp_env_unoptshared/bin/python3 pyston/run_profile_task.py
Error:
cycles:u: PMU Hardware doesn't support sampling/overflow-interrupts. Try 'perf stat'
```

undingen (changes requested):
Needs changes unopt build is also running bolt on the .so

PR #79, opened 2021-08-30
Not sure about this: configure our submodules to be shallow

### LINKED ISSUE #75: Repo size - Submodules / Parallel Building Memory Usage
Two issues:

1. With all the submodules, the git pack files *alone* are 3.7GiB and with everything checked out it's in the neighborhood of 6 GiB... It would be nice to try and slim that down if possible. (Can git submodules be set to do a shallow fetch?)

2. Building in parallel (```make -j```) on a VM with 32 cores exposed (the physical hardware is 2x Xeon 8-core with hyperthreading, so 16 threads/CPU) results in ~64 parallel jobs running - at least in the llvm/bolt phase. The trouble with this is:
  * Massive memory usage (something like 20GB)
  * Probably slower build times (depending on what I'm building, I find that 24-30 jobs is actually fastest ... because the hyperthreading isn't actually another core.)
  * The normal make options like ```-j16``` aren't respected - when I try this, bolt and llvm appear to build in parallel but using one thread each. 
  * I haven't tested the I/O impact specifically, but using all the cores can definitely run into interesting situations where the build is I/O limited and not processor limited. (I'm not using NVMe drives in this case...)

kmod:
Fixes #75 issue 1

There are some drawbacks to this so I don't know if we should merge this, but it was easy to change so I'm putting it up.

With the change:
- the .git folder goes from 3.8GB to 3.2GB
- the total checkout goes from 6.1GB to 5.5GB
- the initial clone takes much longer. I didn't time it but I estimate it went from about 1 minute to 5 minutes. I think this is because of the extra load on the server, where shallow clones don't benefit from the same optimizations that a full clone receives.
- It'd be annoying to change the submodules, though we rarely do that

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,33 +1,43 @@
 [submodule "pyston/bolt/bolt"]
 	path = pyston/bolt/bolt
 	url = https://github.com/facebookincubator/BOLT
+	shallow = true
 [submodule "pyston/LuaJIT"]
 	path = pyston/LuaJIT
 	url = https://github.com/LuaJIT/LuaJIT.git
+	shallow = true
 [submodule "pyston/llvm"]
 	path = pyston/llvm
 	url = https://github.com/llvm/llvm-project.git
+	shallow = true
 [submodule "pyston/tools/FlameGraph"]
 	path = pyston/tools/FlameGraph
 	url = https://github.com/brendangregg/FlameGraph
+	shallow = true
 [submodule "pyston/macrobenchmarks"]
 	path = pyston/macrobenchmarks
 	url = https://github.com/pyston/python-macrobenchmarks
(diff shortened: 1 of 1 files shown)


undingen (commented):
I don't know enough about git so really decide if this a good idea or not.
But in general I sounds like a good trade-off to me (less disk usage and network bandwidth). 
How will the workflow for updating a submodule for us change? I guess we will have to do a full checkout but is something else necessary?

kmod:
I think I'm -1 on this change mostly because of the clone-time increase and that it's only a 20% improvement in disk space, but I could be convinced

As for workflow, I think you can revert this commit and then when you update your submodules they will convert back to full clones, then you can update them, and then presumably people with shallow clones will get updated to a new shallow clone.

PR #80, opened 2021-08-30
Switch our cmake projects from ninja to make

### LINKED ISSUE #75: Repo size - Submodules / Parallel Building Memory Usage

kmod:
Ninja is nicer but make has the key feature of integrating with the jobserver
of the top-level Makefile

Fixes #75 part 2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,28 +26,28 @@ tune: pyston/build/system_env/bin/python
 tune_reset: pyston/build/system_env/bin/python
 	PYTHONPATH=pyston/tools pyston/build/system_env/bin/python -c "import tune; tune.untune()"
 
-pyston/build/Release/build.ninja:
+pyston/build/Release/Makefile:
 	mkdir -p pyston/build/Release
 	@# Use gold linker since ld 2.32 (Ubuntu 19.04) is unable to link compiler-rt:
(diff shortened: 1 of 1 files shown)


undingen (approved):
LGTM

PR #81, opened 2021-08-31
Regenerate .pyc files after package installation

### LINKED ISSUE #70: .pyc files included in the v2.3 deb files have the wrong timestamp
In at least the `pyston_2.3_20.04.deb` file (I didn't check the others) from the recently released [v2.3](https://github.com/pyston/pyston/releases/tag/v2.3), the timestamps stored inside the `.pyc` files do not match the last-modified times of the corresponding `.py` files. As a result, either:

 * If running pyston as non-root, the `.pyc` files are not used and the `.py` files are recompiled each time.
 * If running as root, the `.pyc` files are replaced (after which they are used normally).

Perhaps something in the build process changed the timestamps of all of the `.py` files after they were compiled into `.pyc`s?

As an example, `secrets.py` has a last-modified time of 2021-07-08 15:46:12 UTC (0x60e71dc4 past the epoch), however `__pycache__/secrets.pyston-23.pyc` includes the timestamp 0x60f0ad96 (inside the file at offset 8 in little-endian format) which apparently was the last-modified time of `secret.py` when the `.pyc` file was compiled (2021-07-15 21:50:14 UTC).

undingen:
The files contain the full path of installation directory used during the build
e.g. /home/marius/dev/pyston_v2/pyston/build/cpython_unopt_install/usr/lib/python3.8-pyston2.3/re.py
and as pointed out in #70 also the wrong timestamp.
This means the common use-case where the package is installed by root and a different user is executing pyston
can't use the .pyc files and instead has to parse the files on every import.

The ubuntu cpython package contains a similar script but which does not use `compileall.py` but instead
manually does most of the work and is not regenerating them for all optimization levels.
```bash
files=$(dpkg -L lib@PVER@-testsuite | sed -n '/^\/usr\/lib\/@PVER@\/.*\.py$/p' | egrep -v '/lib2to3/tests/data|/test/bad')
    if [ -n "$files" ]; then
        /usr/bin/@PVER@ -E -S /usr/lib/@PVER@/py_compile.py $files
        if grep -sq '^byte-compile[^#]*optimize' /etc/python/debian_config; then
            /usr/bin/@PVER@ -E -S -O /usr/lib/@PVER@/py_compile.py $files
        fi
    else
        echo >&2 "@PVER@: can't get files for byte-compilation"
    fi
```
I decided against copying the code and instead decided to stick to what cpython does in `Makefile.pre.in`
using `compileall.py`.

This does not change things for the portable package but I think here it's not a problem
because the directory will likely be writable by the user so pyston will just update
them the first time they are imported.

In addition added a script to remove the .pyc files when removing the package because dpkg will only remove the
files included in the package and will complain that it can't remove the directories if they contain newly created .pyc files.

Note: the files to exclude (e.g bad_coding,..) is taken from `Makefile.pre.in`. This is necessary because some test files contain stuff which would error.

Fixes #70

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ make script_opt
 4. adjust `PYSTON_VERSION` inside `configure.ac` and run `autoconf`
 5. update `PYSTON_MINOR` and similar inside Makefile
 6. update the include directory in pyston/pystol/CMakeLists.txt
-7. update pyston/debian/pyston.{install,links}
+7. update pyston/debian/pyston.{install,links,postinst,prerm}
 
 ## Release packaging
 We use a script which builds automatically packages for all supported distributions via docker (will take several hours):
--- a/pyston/debian/pyston.postinst
+++ b/pyston/debian/pyston.postinst
@@ -1,0 +1,20 @@
+#!/bin/sh
+set -eu
+
+PYSTONLIBPATH=/usr/lib/python3.8-pyston2.3
+
+if [ "$1" = configure ]; then
+    # regenerate the .pyc files because they include incorrect timestamps and filepaths
+    # in addition pregenerate them also for the different optimization levels.
+    for OPT in "" "-O" "-OO"
+    do
+        /usr/bin/pyston -E -S $OPT $PYSTONLIBPATH/compileall.py \
+                        $PYSTONLIBPATH/ \
+                        -j0 -f -q \
+                        -x 'bad_coding|badsyntax|site-packages|lib2to3/tests/data'
+        /usr/bin/pyston -E -S $OPT $PYSTONLIBPATH/compileall.py \
+                        $PYSTONLIBPATH/site-packages \
+                        -j0 -f -q \
+                        -x badsyntax
+    done
+fi
--- a/pyston/debian/pyston.prerm
+++ b/pyston/debian/pyston.prerm
@@ -1,0 +1,11 @@
+#!/bin/sh
+set -eu
+
+PYSTONLIBPATH=/usr/lib/python3.8-pyston2.3
+
+if [ "$1" = remove ]; then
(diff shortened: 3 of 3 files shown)


kmod:
LGTM! Thanks for digging into this

gurnec:
You'll probably want to limit what is deleted from `site-packages` to something along these lines:

```shell
PYSTONLIBPATH=/usr/lib/python3.8-pyston2.3
PYSTONSITEPATH=$PYSTONLIBPATH/site-packages

if [ "$1" = remove ]; then
    # remove .pyc files else we get these warnings:
    # dpkg: warning: while removing pyston, directory '<dir>/__pycache__' not empty so not removed
    find $PYSTONLIBPATH -name "*.pyc" ! -path "$PYSTONSITEPATH/*" -delete
    find $PYSTONSITEPATH/pkg_resources $PYSTONSITEPATH/pip $PYSTONSITEPATH/setuptools -name "*.pyc" -delete
    rm -f $PYSTONSITEPATH/__pycache__/easy_install.pyston-23.*pyc
fi
```

Otherwise you may and up deleting `.pyc`s from admin-installed libraries.

Likewise, it may make sense to be restrictive during install:

```shell
        /usr/bin/pyston -E -S $OPT $PYSTONLIBPATH/compileall.py \
                        $PYSTONSITEPATH/pkg_resources \
                        $PYSTONSITEPATH/pip \
                        $PYSTONSITEPATH/setuptools \
                        $PYSTONSITEPATH/easy_install.py \
                        -j0 -f -q \
                        -x badsyntax
```

Just a thought... (fyi this is untested)

undingen:
@gurnec good point I put up a new patch which should fix the issues:
https://github.com/pyston/pyston/pull/84

Issue #76, opened 2021-08-08
Can Pyston run uvicorn server?


EmadMokhtar:
I'm building a Web API using FastAPI framework. It used `uvicorn` as a server to run the application. Can Pyston run uvicorn server? I'm not sure because I'm getting this error.

``` bash
uvicorn main:app --reload                                                                                        
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [31274] using watchgod
Traceback (most recent call last):
  File "/home/emadmokhtar/.pyenv/versions/fastapi-pyston/bin/uvicorn", line 8, in <module>
    sys.exit(main())
  File "/home/emadmokhtar/.pyenv/versions/pyston-2.2/envs/fastapi-pyston/lib/pyston3.8/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/home/emadmokhtar/.pyenv/versions/pyston-2.2/envs/fastapi-pyston/lib/pyston3.8/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/home/emadmokhtar/.pyenv/versions/pyston-2.2/envs/fastapi-pyston/lib/pyston3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/emadmokhtar/.pyenv/versions/pyston-2.2/envs/fastapi-pyston/lib/pyston3.8/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/home/emadmokhtar/.pyenv/versions/pyston-2.2/envs/fastapi-pyston/lib/pyston3.8/site-packages/uvicorn/main.py", line 371, in main
    run(app, **kwargs)
  File "/home/emadmokhtar/.pyenv/versions/pyston-2.2/envs/fastapi-pyston/lib/pyston3.8/site-packages/uvicorn/main.py", line 388, in run
    ChangeReload(config, target=server.run, sockets=[sock]).run()
  File "/home/emadmokhtar/.pyenv/versions/pyston-2.2/envs/fastapi-pyston/lib/pyston3.8/site-packages/uvicorn/supervisors/basereload.py", line 43, in run
    self.startup()
[...]

Issue #41, opened 2021-05-22
Pyston can't get freetype2 version because ZLIB 1.2.9 is not found


rocky:
I've been trying pyston 2.2 for running an open-source Mathematica called [Mathics](https://mathics.org)

Using it is about 25-30% faster than CPython as advertised. Thanks!

In doing this and running package [wordcloud](https://pypi.org/project/wordcloud/) 

I get this traceback in part...

```
  File "/src/external-vcs/github/mathics/Mathics/mathics/builtin/drawing/image.py", line 2560, in _word_cloud
    wc.generate_from_frequencies(freq)
  File "/home/rocky/.pyenv/versions/pyston-2.2/lib/pyston3.8/site-packages/wordcloud/wordcloud.py", line 496, in generate_from_frequencies
    font = ImageFont.truetype(self.font_path, font_size)
  File "/home/rocky/.pyenv/versions/pyston-2.2/lib/pyston3.8/site-packages/PIL/ImageFont.py", line 853, in truetype
    return freetype(font)
  File "/home/rocky/.pyenv/versions/pyston-2.2/lib/pyston3.8/site-packages/PIL/ImageFont.py", line 850, in freetype
    return FreeTypeFont(font, size, index, encoding, layout_engine)
  File "/home/rocky/.pyenv/versions/pyston-2.2/lib/pyston3.8/site-packages/PIL/ImageFont.py", line 173, in __init__
    freetype_version = parse_version(features.version_module("freetype2"))
  File "/home/rocky/.pyenv/versions/pyston-2.2/lib/pyston3.8/site-packages/packaging/version.py", line 57, in parse
    return Version(version)
  File "/home/rocky/.pyenv/versions/pyston-2.2/lib/pyston3.8/site-packages/packaging/version.py", line 296, in __init__
    match = self._regex.search(version)
```

What is going on above is that `version` from the call `parse_version(...("freetype2")` is coming back `None` and that is because the dynamic loader is not finding`ZLIB_1.2.9`:

```
 __import__('PIL._imagingft')
** ImportError: /home/rocky/.pyenv/versions/pyston-2.2/lib/pyston3.8/lib-dynload/../../../lib/libz.so.1: version `ZLIB_1.2.9' not found (required by /lib/x86_64-linux-gnu/libpng16.so.16)
```

kmod:
Thanks for the report! Did you install from the portable build or the debian package?

Also if you could provide a way to reproduce this error that would help us look into it.

[...]

PR #83, opened 2021-09-02
portable release: prefer distro supplied shared objects

undingen:
Until now we always used our supplied 3rd party shared objects dependencies over
the distro supplied ones. But ours are likely older which causes problems.
E.g. issue #41: we load a thirdparty distro library compiled against a newer zlib but it's loading
our supplied older version which fails.

This patch changes the search order to use our supplied .so's only as fallback if the dist ones are not available.

This is noticable e.g. running `pyston -c "import sqlite3; print(sqlite3.sqlite_version)"`
will output 3.31.1 (on my ubuntu 20.04 system) instead of 3.11.0.

Why do we even supply this libraries?:
We do have to supply this libraries in case the user does not have the library installed
or it's not available in the required major version. E.g. if you use the portable release on latest fedora and ubuntu
openssl 1.0.x is not available anymore (it uses 1.1.x) so pip will fail with a SSL not available error if we don't supply 1.0.
Which means in order to have only one release which works on as much distros as possible we need to supply this
files but should make sure they are only used as fallback.

I verified that this fixes the problems in #41 and #76 on ubuntu, fedora and archlinux.

BTW: patchelf v0.10 supplied in ubuntu 20.04 contains a bug which corrupts files when modifying
already previously via patchelf modified executables. Not an issue for this script because it's not running patchelf twice on the same file but something to lookout if manually changing rpaths. Newer and older version seem to work fine.

### DIFF
--- a/pyston/tools/make_portable_dir.py
+++ b/pyston/tools/make_portable_dir.py
@@ -114,7 +114,11 @@ def make_portable_dir(outdir):
 
     for f in filter(is_so, os.listdir(path)):
         dependencies |= recursive_get_dependencies(Dependency(f, path + f))
-        set_rpath(path + f, "$ORIGIN/../../../lib")
+        # we like to use our supplied .so's only as fallback if the dist ones are not available
+        # because ours are likely older which causes problems.
+        # best way to do it seems to put the most common dist lib paths in front of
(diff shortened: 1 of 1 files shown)


kmod:
:+1:

PR #84, opened 2021-09-07
Regenerate .pyc files: only process files part of the package

undingen:
Before packages installed by a user inside the 'site-packages' directory also got processed.

The pyston.prerm: script lines are mostly copied from the ubuntu package.


kmod:
:+1:

Issue #77, opened 2021-08-14
Segmentation fault while using python only code along the official binary without C extension


ytrezq:
Install [panoramix](https://github.com/palkeo/panoramix)

run `/usr/bin/panoramix "0xe0090Ec6895c087a393F0e45f1F85098A6c33bEf"`
The process crashes about 2 hours laters just before completion.

kmod:
Sorry for the delay, we're on vacation and should check this out soon

undingen:
Thanks for letting us know about this crash :).

I could reproduce the issue [1]. The cause is that decompiling `safeApprove` should raise a `RecursionError` but we removed the check in 2.3.
```
11:57:49 panoramix.decompiler Problem with safeApprove(address _param1, address _param2, uint256 _param3)
Traceback (most recent call last):
  File "/home/marius/dev/pyston_v2/pyston/build/unopt_env/lib/python3.8-pyston2.3/site-packages/panoramix/decompiler.py", line 177, in _decompile_with_loader
    trace = dec()
  File "/home/marius/dev/pyston_v2/pyston/build/unopt_env/lib/python3.8-pyston2.3/site-packages/timeout_decorator/timeout_decorator.py", line 82, in new_function
    return function(*args, **kwargs)
  File "/home/marius/dev/pyston_v2/pyston/build/unopt_env/lib/python3.8-pyston2.3/site-packages/panoramix/decompiler.py", line 169, in dec
    trace = make_whiles(trace)
  File "/home/marius/dev/pyston_v2/pyston/build/unopt_env/lib/python3.8-pyston2.3/site-packages/panoramix/whiles.py", line 99, in make_whiles
    trace = simplify_trace(trace)
  File "/home/marius/dev/pyston_v2/pyston/build/unopt_env/lib/python3.8-pyston2.3/site-packages/panoramix/simplify.py", line 176, in simplify_trace
    trace = cleanup_conds(trace)
  File "/home/marius/dev/pyston_v2/pyston/build/unopt_env/lib/python3.8-pyston2.3/site-packages/panoramix/simplify.py", line 1215, in cleanup_conds
    if_true = cleanup_conds(if_true)
  File "/home/marius/dev/pyston_v2/pyston/build/unopt_env/lib/python3.8-pyston2.3/site-packages/panoramix/simplify.py", line 1215, in cleanup_conds
    if_true = cleanup_conds(if_true)
[...]

PR #85, opened 2021-09-08
Add back recursion checking, but do it faster using the C stack pointer

kmod:
We removed recursion checking because it was about 0.8% of the runtime of our macrobenchmarks, and presumably more in microbenchmarks. The user experience is not very good though, because if there is an infinite recursion the result is a segfault which looks like our fault.

So this PR adds back recursion checking, but instead of explicitly counting Python frames, it just looks at the C stack pointer. This means that we don't have to do any additional writes to do this bookkeeping. This fixes issues like #77.

Additionally, backport an optimization from newer versions of CPython where they eliminated some of the ability to temporarily exceed the recursion limit, which further speeds things up.

The numbers are small enough to be hard to measure, but it looks like CPython-style recursion checking takes up roughly 0.8% of the time, whereas this rsp-based checking takes up about 0.2%.

### DIFF
--- a/Include/ceval.h
+++ b/Include/ceval.h
@@ -83,45 +83,41 @@ PyAPI_FUNC(int) Py_MakePendingCalls(void);
    http://mail.python.org/pipermail/python-dev/2008-August/082106.html
    for some observations.
 */
+#if !PYSTON_SPEEDUPS
 PyAPI_FUNC(void) Py_SetRecursionLimit(int);
 PyAPI_FUNC(int) Py_GetRecursionLimit(void);
+#endif
 
 #define Py_EnterRecursiveCall(where)  \
-            (_Py_MakeRecCheck(PyThreadState_GET()->recursion_depth) &&  \
+            (_Py_MakeRecCheck(PyThreadState_GET()->stack_limit) &&  \
              _Py_CheckRecursiveCall(where))
 #define Py_LeaveRecursiveCall()                         \
-    do{ if(_Py_MakeEndRecCheck(PyThreadState_GET()->recursion_depth))  \
+    do{ if(_Py_MakeEndRecCheck((void*)(((char*)PyThreadState_GET()->stack_limit) + (1<<16))))  \
       PyThreadState_GET()->overflowed = 0;  \
     } while(0)
 PyAPI_FUNC(int) _Py_CheckRecursiveCall(const char *where);
 
+#if !PYSTON_SPEEDUPS
 /* Due to the macros in which it's used, _Py_CheckRecursionLimit is in
    the stable ABI.  It should be removed therefrom when possible.
 */
 PyAPI_DATA(int) _Py_CheckRecursionLimit;
 PyAPI_DATA(int) _Py_RecursionLimitLowerWaterMark_Precomputed;
+#endif
 
 #ifdef USE_STACKCHECK
 /* With USE_STACKCHECK, trigger stack checks in _Py_CheckRecursiveCall()
    on every 64th call to Py_EnterRecursiveCall.
 */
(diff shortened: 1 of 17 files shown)


kmod:
Whoops I didn't run the tests, don't look at this just yet

kmod:
Ok I committed a fix and this is now ready to review

Include/ceval.h
```diff
@@ -86,48 +86,39 @@ PyAPI_FUNC(int) Py_MakePendingCalls(void);
...
+#define frame_address() _stack_pointer()
 
 #ifdef USE_STACKCHECK
 /* With USE_STACKCHECK, trigger stack checks in _Py_CheckRecursiveCall()
    on every 64th call to Py_EnterRecursiveCall.
 */
-#  define _Py_MakeRecCheck(x)  \
-    (++(x) > _Py_CheckRecursionLimit || \
+#  define _Py_MakeRecCheck()  \
+    (frame_address() < (x) || \
```
undingen, line 116: I think this code will not work because of the removal of `x` argument.
I'm also fine with just raising a explicit compile error if USE_STACKCHECK is defined but we have not tested it.

Python/ceval.c
```diff
@@ -667,9 +665,27 @@ void
...
+
+    _Py_AdjustThreadStackLimits(new_limit - old_limit);
+}
+
+// The actual value for a Python frame calling another Python frame seems to be about 500 bytes,
+// but set the value to something higher to guarantee that you get at least as many frames as
+// you request.
+#define BYTES_PER_RECURSION_LEVEL 800
+void* _Py_GetStackLimit(int levels) {
+    return (char*)__builtin_frame_address(0) - BYTES_PER_RECURSION_LEVEL * levels;
```
undingen: I think this can be `frame_address()`

pyston/nitrous/Execution.cpp
```diff
@@ -1175,6 +1176,16 @@ void Interpreter::visitCallSite(CallSite CS) {
     case Intrinsic::dbg_value:
       return;
 
+    case Intrinsic::frameaddress: {
```
undingen, line 1179: I think we should be able to remove this again?!?
I worry a bit that w could run into some code which is using this in the future and we silently do something really strange while tracing it instead of obvious error code.

pyston/test/test_recursion_limit.py
```diff
@@ -0,0 +1,43 @@
+import sys
```
undingen, line 1: I think it makes sense to add this tests in the `Makefile` to the `_runtests` target.

pyston/test/test_recursion_limit.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # Basic test: check that setting the limit is close
+    sys.setrecursionlimit(set_limit)
+    actual_limit = measure()
+    assert set_limit - 3 <= actual_limit < 2 * set_limit, actual_limit
+
+    # Test that doubling the limit roughly doubles the number of frames
+    sys.setrecursionlimit(set_limit * 2)
+    actual_limit2 = measure()
+    assert 2 * actual_limit - 1 <= actual_limit2 <= 2.1 * actual_limit + 2, (actual_limit, actual_limit2)
```
undingen: Ok I have just noticed that depending on the value of  `JIT_MIN_RUNS` this assert fails. I guess it's a bit too specific `assert 1.9 * acutal_limit` makes it work for me an all thresholds I tried

undingen (commented):
Very nice approach :+1: added some minor comments.

kmod:
Thanks for the catches!

I'm cleaning up the test running in my next PR, including automatically running all of the tests in pyston/test

PR #86, opened 2021-09-08
Merge in upstream 3.8.12

kmod:
I noticed that the test_ssl test had started failing; it's failing on 3.8.8 in upstream CPython but they fixed it for 3.8.12. So merge up to their latest 3.8 release.

The merge was clean except for some files that I deleted for our repo (github actions, their readme), and I checked that at least the cpython testsuite passes.


undingen (approved):
LGTM, merging all the bug and security fixes sounds like a very good idea.

PR #88, opened 2021-09-13
Fix the tests

kmod:
Main thing was that the merge of v3.8.12 changed the behavior of the django testsuite. One option would have been to make sure to compare to a v3.8.12 cpython, but I updated django and the issue is fixed.

I had another test failure but it was due to a bad wheel in the pip cache.
But in the process of debugging that I noticed that our debug build wasn't working, so I have one makefile fix and one jit fix for the debug build.

I cleaned up and improved the makefile a little bit too -- all of the .py files in pyston/test/ are now automatically included in our test runs.


Python/aot_ceval_jit.c
```diff
@@ -729,7 +729,7 @@ static void emit_dec_qword_ptr(Jit* Dst, void* ptr, int can_use_tmp_reg) {
 }
 static void emit_inc_qword_ptr(Jit* Dst, void* ptr, int can_use_tmp_reg) {
     if (IS_32BIT_VAL(ptr)) {
-        | inc qword [&ptr]
+        | inc qword [ptr]
```
undingen, line 732: nice catch!
this is easily reproducible: `SHOW_JIT_STATS=1 make nbody_unopt` crashed before don't know how I missed that... :(

undingen (approved):
LGTM thanks for catching the jit bug :+1:
I run into a pyston test failure which I fixed: https://github.com/pyston/pyston/pull/93
also test_numpy is still not working for me because of additional skipped tests everything else seems to succeed.

kmod:
Ah yeah I only ran `make _runtests`

PR #89, opened 2021-09-14
Make a bunch of names shorter so we can do less typing

kmod:
pyston/build/cpython_bc_build => build/bc_build
pyston/bolt/bolt => pyston/bolt
call_function_ceval_no_kwProfile => call_functionProfile


undingen (changes requested):
LGTM with this small fixes: 
- `pyston/debian/rules` and `pyston/debian/pyston.install` still contain the old paths
- `README`: `git submodule update --init pyston/llvm pyston/bolt/bolt pyston/LuaJIT pyston/macrobenchmarks` has wrong dir to `bolt` and also contains a reference to `cpython_unopt_install`
- `pyston/tools/make_src_release.sh` also contains wrong `bolt` path

undingen:
`.dockerignore` also needs a change

kmod:
Whoops, thanks for the catches, I think I caught things now

PR #91, opened 2021-09-14
Get embedding (ex uwsgi) working from our portable release

kmod:
The problem is that sysconfig.get_config_var("LIBDIR") returns
the configured path, but the release is most likely extracted
somewhere else.

So at sysconfig-load time, rewrite LIBDIR to always point to
where we think we're running from.

I'm a bit nervous about this change because it changes the
semantics of LIBDIR. I think it's for the better, but there's
a chance that someone really wants to look at the configured
location, or it's possible that there's some weird situation
where we incorrectly guess where we're running from.

There are also a number of other variables that are like this,
that could be updated to point to the location we think we're
running from, but I'm doing this just for LIBDIR for now
to reduce the amount of changes.


undingen (approved):
Great! I tested the new portable release on fedora and could build and run uwsgi :100:

PR #93, opened 2021-09-15
jit: don't crash if mmap fails

undingen:
this fixes `JIT_MIN_RUNS=0 make pyston/test/test_code_dealloc_unopt`


kmod:
:+1:

why does mmap fail? I would have thought the memory limit would save us here

undingen:
The test contains `resource.setrlimit(resource.RLIMIT_AS, (2<<20, 2<<20))` ~2MB and we hit this virtual memory limit after
generating about 1.23MB of jitted code well below our 100MB jit limit.

PR #94, opened 2021-09-16
pgo: run profile tasks in parallel

undingen:
I could not measure any performance difference of the pyston executable
but the opt build takes about 10min now instead of 25min+ or so.

I could not really find any concrete sentence that says multiprocess PGO is fine but the docs mention
that the profile counters are merged on exit. Based on the benchmark result I assume it merges
them locked.


kmod:
ok cool, glad it just works!

PR #95, opened 2021-09-16
wip: change bolt .so strategy back to instrumentation

kmod:
They said the issue was fixed of using instrumentation on a shared library, so I updated bolt and set it back to using instrumentation.
Unfortunately we see some sort of behavioral change and it throws an exception during interpreter initialization.

I didn't debug it since it's not urgent for us to look into, but I thought I would put up the code so that we could pick it back up later.


undingen:
thanks for trying! I was also wandering if it would work now...

undingen:
I spend some time on the weekend to figure out whats going on, I think I know now a bit more:
It seems like the issue is caused by `-jump-tables=move` causing issues on some functions which we already run into before but worked around by setting `-jump-tables=none`. But `-instrument` sets `-jump-tables=move` internally so we run into the same issue again. Adding `-skip-funcs=_PyEval_EvalFrameDefault -skip-funcs=_PyEval_EvalFrame_AOT_Interpreter/1` fixes the issue for me. But I have not tested if the instrumentation works / same speed as using perf.

PR #96, opened 2021-09-17
Fix compiler warnings

undingen:
This actually caught a bug where in the jit cache stats where load method misses
would get count as load attribute misses...


kmod:
:+1:

PR #97, opened 2021-09-17
portable release: put files into pyston_2.3 directory

undingen:
Instead of archiving the files directly put them in a pyston_<version> directory.


kmod:
:+1:

PR #99, opened 2021-09-18
Add the sqlalchemy and pandas testsuites

kmod:
They both pass, though it took a little bit of work to get them working


undingen (approved):
LGTM

Issue #98, opened 2021-09-17
[Enhancement] - Add a base image on Dockerhub


sorasful:
Hello ! 

Since it seems it's a little bit complicated to setup Pyston, I was thinking it may be a good idea to create a base image so people can start from it, saving the time to set it up correctly.

And it may be useful to upload it on Dockerhub for easier distribution.

What do you think ?

undingen:
Good idea, I also recently learned that it's very easy to execute a linux docker image inside macOS so it would also make it easier for people using apple hardware to use/try out pyston.

kmod:
Hi @sorasful! I just uploaded an image to Dockerhub under pyston:pyston. I don't use docker much myself, would you mind giving it a try?

sorasful:
Hey @kmod , I put a little comment on your commit. But to me that seems cool for testing purposes. For production image, we may need to do some cleanup to have a smaller image !

kmod:
There's stuff that could be improved, but we have the image now! So I'm closing this

PR #101, opened 2021-09-22
Produce a pyston-based docker container

kmod:
In theory you could replace "FROM ubuntu:20.04" with "FROM pyston:pyston",
but since I don't have any docker workloads this will need some testing.

Implements #98

I already uploaded the 2.3 release to pyston:pyston


undingen (approved):
LGTM.

PR #104, opened 2021-09-24
conda packages

undingen:
__Needs changes:__
- [x] replace 'python'
- [x] remove pip (should be own package)
- [x] figure out how to package all the 3rd party libraries (python abi string)
- [ ] fix package dependencies - are they correct? compare with cpython dependencies
- [ ] do all the builtin modules get build? compare with cpython conda package
- [ ] check if sysconfig is correct
- [ ] run tests during build - maybe we need to remove the tests from the final package
- [ ] strip binaries - I think they are not stripped now that we don't use dpkg to build
- [ ] generate correct .pyc files
- [ ] switch to 'release' from 'opt' build - maybe blocked by instrumentation problems in bolt
- [x] fix test failures during profiling ("file was modified... default.profraw")

__May need changes:__
- [ ] we internally pip install packages into a temporary environment for the bolt profiling task. I think this is okay because this package will not end up in the final conda package and we install them via exact version string but may still be preferred to supply the source packages locally instead.
- [x] ~bolt pkg directly loads source archive from github because we can only checkout tags. Maybe we should fork it and create tags ourselves instead. I named the bolt release we are using after the date of the last commit~ Forked bolt into the pyston org and created a tag which we point to.
 
__How to try it out:__
run the following script to build our packages in a docker miniconda image which outputs the finished packages into `release/conda_pkgs/`
```
$ pyston/conda/build_pkgs.sh
```
Afterwards index the directory and create a activate new enviroment called pyston_env
```
$ conda index -c pyston release/conda_pkgs/
$ conda create -n pyston_env pyston -c `pwd`/release/conda_pkgs/ -c conda-forge --override-channels
$ conda activate pyston_env
```
Now one can e.g install attrs package:
```
(pyston_env)$ conda install --override-channels -c conda-forge -c `pwd`/release/conda_pkgs/ attrs
(pyston_env)$ python -c "import attr"
```
Numpy:
```
(pyston_env)$ conda install --override-channels -c conda-forge -c `pwd`/release/conda_pkgs/ numpy
(pyston_env)$ python -c "import numpy as np; print(np.identity(5))"
```
__Note:__
I used conda-forge instead of the default anaconda channel because conda-forge packages have the python_abi tag set -
on the main channel it will happily let you install cpython extensions compiled against cpython.

__Note:__
The `compiler-rt` recipe is from anaconda but they currently don't build a linux_64 version thats why we also have compile it for now and I included it.

### DIFF
--- a/pyston/conda/build_pkgs.sh
+++ b/pyston/conda/build_pkgs.sh
@@ -30,6 +30,12 @@ do
     CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY=0 conda build \${pkg}-feedstock/recipe --python="${PYSTON_PKG_VER}" --override-channels -c conda-forge --use-local
 done
 
+# build numpy 1.20.3 using openblas
+git clone https://github.com/AnacondaRecipes/numpy-feedstock.git -b pbs_1.20.3_20210520T162213
+# 'test_for_reference_leak' fails for pyston - disable it
+sed -i 's/_not_a_real_test/test_for_reference_leak/g' numpy-feedstock/recipe/meta.yaml
+conda build numpy-feedstock/ --python="${PYSTON_PKG_VER}" --override-channels -c conda-forge --use-local --extra-deps pyston --variants="{blas_impl: openblas, openblas: 0.3.3, c_compiler_version: 7.5.0, cxx_compiler_version: 7.5.0}"
+
 for arch in noarch linux-64
 do
     mkdir /conda_pkgs/\${arch}
--- a/pyston/conda/build_pkgs.sh
+++ b/pyston/conda/build_pkgs.sh
@@ -2,11 +2,16 @@
 set -eux
 
 PYSTON_PKG_VER="3.8.5 *_23_pyston"
-OUT_DIR=${PWD}/release/conda_pkgs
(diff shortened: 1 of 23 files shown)


pyston/conda/build_pkgs.sh
```diff
@@ -0,0 +1,23 @@
...
+
+docker run -iv${PWD}:/host-volume --rm continuumio/miniconda3 sh -s <<EOF
+set -ex
+conda install conda-build -y
+conda build pyston/conda/compiler-rt
+conda build pyston/conda/bolt
+conda build pyston/conda/pyston
+conda install pyston --use-local
+
+wget
```
kmod: Is this a leftover?

kmod (commented):
This is looking great!

pyston/conda/build_pkgs.sh
undingen: I accidentally added that file :blush:.

undingen:
I think this is ready for a review. I updated the instructions how to run it and also packaged 'numpy'.
There are still things to improve (items in the todo list) but I think we can fix it in new PRs as now the basic things seem to work and we can start packaging 3rd party libraries.

kmod:
Cool this looks great!

PR #107, opened 2021-09-28
Get rid of the usr/ directory in our portable release

kmod:
Seems unnecessary now that we have a pyston-2.3.1 top level directory

I already put this up on our release as the pyston-2.3.1-portable-v2.tar.gz


undingen (approved):
LGTM

PR #108, opened 2021-10-05
setup.py: Fix typos discovered by codespell

cclauss:
https://pypi.org/project/codespell


kmod:
Good catches! These are positive improvements, but given our circumstance of being a fork of another project I don't believe it's the best idea to make these changes here. Looks like the typos have been fixed upstream so we'll get them when we update.

PR #109, opened 2021-10-06
Improve Dockerfile

PLPeeters:
* Reduce image size by cleaning the package list after running apt-get (341 -> 309 MB)
* Use variables and symlink existing symlinks to reduce diff size of future updates
* Add `python` symlink

### DIFF
--- a/pyston/docker/Dockerfile
+++ b/pyston/docker/Dockerfile
@@ -1,16 +1,21 @@
 FROM ubuntu:20.04
 
-RUN apt-get update
-RUN apt-get install -y wget
+ENV PYSTON_VERSION='2.3.1'
+ENV UBUNTU_VERSION='20.04'
 
 RUN set -eux; \
-    wget https://github.com/pyston/pyston/releases/download/pyston_2.3.1/pyston_2.3.1_20.04.deb; \
-    apt install -y ./pyston_2.3.1_20.04.deb; \
-    rm pyston_2.3.1_20.04.deb
+    apt-get update; \
+    apt-get install -y wget; \
+    wget https://github.com/pyston/pyston/releases/download/pyston_${PYSTON_VERSION}/pyston_${PYSTON_VERSION}_${UBUNTU_VERSION}.deb; \
+    apt install -y ./pyston_${PYSTON_VERSION}_${UBUNTU_VERSION}.deb; \
+    rm pyston_${PYSTON_VERSION}_${UBUNTU_VERSION}.deb; \
+    rm -rf /var/lib/apt/lists/*
 
 RUN set -eux; \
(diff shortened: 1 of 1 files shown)


PLPeeters:
Side-note: the `python3.8` image is actually 909 MB and includes quite a lot of packages that ensure that most `pip install`s will just work. I'm wondering whether this image should go that way as well?

undingen:
LGTM - thanks for the improving our docker image!
@kmod is building our docker image so I think he should review it.

kmod:
lgtm! I uploaded new images from this pr

kmod:
I don't know much about how people use docker for Python, do you think people use python3.8 or something more stripped down like alpine?

Personally I would rather err on the side of having a big image that "just works" than a smaller one that might end up not being a straight upgrade, since I think the target of this is people who are quickly experimenting by switching their docker images to be based on our image. So I think that'd be a great idea to install whatever stuff python3.8 does!

PLPeeters:
@kmod We used the full image for simplicity because it quickly gets annoying to figure out which packages you're missing when trying to install library after library. It still happens of course, just way less often.

The `python3.8` [Dockerfile](https://github.com/docker-library/python/blob/d65295b73bc2c711de96caeba17501d1a81b6a5c/3.8/bullseye/Dockerfile) stems from `buildback-deps:bullseye` and basically adds the `libbluetooth-dev`, `tk-dev` and `uuid-dev` packages. I suppose just doing the same could be enough. Want me to give it a shot?

kmod:
That would be great!

kmod:
#112 makes me think we should install more stuff into the docker container by default; maybe we could offer a "slim" version later if people want.

PLPeeters:
@kmod I'll get to that on Monday then. Should I create a specific issue for this for traceability?

kmod:
No rush! And I don't have a strong opinion but I don't think a new issue is necessary.

LiorA1:
> #112 makes me think we should install more stuff into the docker container by default; maybe we could offer a "slim" version later if people want.

It's just that I usually work with Python slim version, so I didn't understand why the pip, didn't work..

kmod:
Which image specifically?

And I think the difference is that normally you don't compile numpy yourself since you download a pre-compiled version, but there aren't any pre-compiled versions available for Pyston (something we're working on changing).

LiorA1:
Maybe its worth mentioned in the [pyston/pyston](https://hub.docker.com/r/pyston/pyston) page ?

(A basic explanation about which os is used, and how to build numpy, matplotlib etc..)

**I'm only a fresh grad, so it wasn't clear to me..

Edit:
The python-3.8:slim and the alpine..

kmod:
Good idea, I'll include an update in my next PR!

PR #110, opened 2021-10-12
POC: integrate the cycle collector with the memory manager

kmod:
Current state: this has most of the pieces required, but is quite a slowdown and still has problems (ex weakrefs will crash).

## Basic idea

Currently, the cycle collector puts all GC-aware heap objects into a big linked list, and then iterates through the list to enumerate objects.

My (unverified) theory is this is slow because
- It scans memory in a random order causing tons of cache misses
- Every object has a different type meaning that the tp_traverse calls can't be predicted

So I added a new approach: I augmented pymalloc (the small-object memory allocator) to keep a bitmap of which objects belong in which generations and called the new system gcmalloc. I also created separate pymalloc "allocators" with the intention that each allocator will only serve a single type.

The result should be that the heap can be scanned very quickly, in sequential order, and calling the same tp_traverse methods in batches.

## Result

The result is about a 30% slowdown that I haven't investigated. There are also still unimplemented features like weakrefs and maybe also destructors.

The main issue is that the GC uses the linked lists for many things other than generation management. In this PR only the very first scan is done via the bitmap, and then it's all done via linked lists afterwards.

I think the solution to this is to create another bitmap for these temporary lists.

But this would mean that the GC would have to differentially handle gcmalloc'd memory from non-gcmalloc'd memory, and currently this is only tracked inside gcmalloc. I was thinking of using another low bit of the gc_prev field (restricting us to 64-bit systems, which could be acceptable) to store whether the memory was gcmalloc'd or not.

I still believe in this approach but there are way more GC features than I gave it credit for which all need to be updated to work with the new system. I think getting this working would be a fair amount of work. I think the 30% slowdown isn't necessarily indicative of the approach but it's an additional thing that would need to be hunted down.

For example weakrefs are currently broken because the GC tracks just-freed objects and iterates through them for weakref functionality. We would either need to leave the gcmalloc'd objects in the just-freed list (and remove them later) or have another way of tracking them to clear their weakrefs.


undingen:
Unfortunate did turned out to be even more complicated than expected :( (I did not follow your code completely would need to spend more time on it).
Maybe if we pick this up again it's worth comparing with https://github.com/colesbury/nogil - maybe using mimalloc heaps would also be useful for our usecase and maybe simplify keeping of the objects.

PR #114, opened 2021-10-18
Trace int and dict and call function with keyword arguments

undingen:
While investigating function call overhead I notices this missing traces.
About 0.5% speedup on make measure and 15-20% on micro-benchmarks which call the traces in a tight loop.


kmod:
Looks good! Maybe I'm being dense though but how did this enable int/long traces?

undingen:
I had a thinko wanted to write int and dict not int and long.
We are now generating `call_functionLong2` (before it was commented out):
image
Also we did not do any traces for `dict` - we are now creating: `call_functionDict1` and `call_functionDict2`.
The other types jut got the no arg version.

kmod:
Oh I was looking at the wrong commit, thanks!

PR #115, opened 2021-10-18
Change Docker image source to buildpack-deps and add slim image

PLPeeters:
Follow-up on #109. 

Image sizes: 

* 931 MB for the buildpack-deps image (vs 909 MB for the `python:3.8` image)
* 310 MB for the bullseye-slim image (vs 122 MB for the `python:3.8-slim` image, but since the unarchived portable Pyston tarball is 261 MB already, the increase definitely stems from there; not sure how to go about reducing this or if it's even possible)

I tried adding an alpine-based image as well, but could not get the portable version of Pyston to run on it. I'd say this is a good enough start and alpine can be investigated if there is demand for it.

### DIFF
--- a/pyston/docker/Dockerfile
+++ b/pyston/docker/Dockerfile
@@ -1,14 +1,20 @@
-FROM ubuntu:20.04
+FROM buildpack-deps:20.04
 
 ENV PYSTON_VERSION='2.3.1'
 ENV UBUNTU_VERSION='20.04'
 
+# http://bugs.python.org/issue19846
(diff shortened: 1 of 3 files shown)


kmod:
This is awesome, thanks!

kmod:
I uploaded the new images to dockerhub

PLPeeters:
@kmod ~Any reason why the slim image suddenly disappeared from Docker Hub?~ Nevermind, just saw they were moved to `pyston/slim`!

kmod:
Ah sorry yeah I checked that in without mentioning anything, I think this is how the python images do it (separate repo) so I changed it

PR #116, opened 2021-10-21
CI: Add github action which builds pyston and runs the cpython testsuite

undingen:
- builds our unopt and the release build
- runs cpython_testsuite (we can later add more testsuites)
- I uploaded the pyston build requirements packages (`bolt` and `compiler-rt`) to our anaconda [pyston repo](https://anaconda.org/pyston/repo)  but with a label `dev` so that I can not yet get accidentally installed. In case we have to fix the packages before releasing them to the public.
- CI run takes about 25m for the unopt build and 1h for the release build (I think the max limit is 6h but not sure if there is a monthly limit)

### DIFF
--- a/.github/workflows/github-actions-ci.yml
+++ b/.github/workflows/github-actions-ci.yml
@@ -1,0 +1,36 @@
+name: GitHub Actions CI
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          - build: unopt
+            PYSTON_UNOPT_BUILD: 1
+          - build: opt
+            PYSTON_UNOPT_BUILD: 0
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


kmod:
looks cool!

undingen:
Ready for review (if the 'opt' build goes green - currently still running)
It still has the `/usr/include` hack in it but I plan to look into how to prevent our build from picking up system libraries in a new PR.

kmod:
Looks awesome! Where does it run the cpython_testsuite?

undingen:
It calls `make cpython_testsuite[_release]` in unopt and in the release build (during build phase not in the conda test section). This means we will now always run the tests even when not building for CI. But I thought this is something we want anyway. 
image
CI Log: https://github.com/pyston/pyston/runs/4023939717?check_suite_focus=true#step:8:30420

kmod:
Ah ok, I wasn't expecting it to be in the build step

PR #117, opened 2021-10-21
Get the conda package passing the cpython tests

kmod:
- add tk libraries
- set prefix to conda prefix so that conda can rewrite it
- sneak in some unrelated documentation


undingen (approved):
LGTM. Thanks for helping with the conda packaging.

PR #118, opened 2021-10-25
aot: merge function and method and trace bound methods

undingen:
The only difference was that function traces assumed that the oparg argument
stays the same while for CALL_METHOD they can change.

This will add one if check which will always be false in exchange
for a reduction in then number of traces.
I think this is a good trade-off. When I added the differentiation between
call_function and call_method I did not realize that e.g. calling a function inside another module will use the CALL_METHOD
that's why I thought it would be useful to have different traces.

Measuring a small speedup on 'make measure'

### DIFF
--- a/Objects/classobject.c
+++ b/Objects/classobject.c
@@ -43,7 +43,7 @@ PyMethod_Self(PyObject *im)
 }
 
 
-static PyObject *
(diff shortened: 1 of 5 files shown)


kmod:
good idea!

PR #119, opened 2021-10-26
jit: introduce new jit for call argument setup

undingen:
this JIT generates code for `CALL_FUNCTION`, `METHOD_CALL` and `CALL_FUNCTION_KW` opcodes.
The idea is to reduce the time spend inside `_PyEval_EvalCodeWithName()` by generating machine code for it.
By guarding on the function version we know that the default args, code object etc can't change.
This allows us to directly emit default args and kwargs (before it had to loop on every call to retrieve position etc..) similar to pyston v1.
With this change the following calls are all the same speed:
```python
def f(a, b, c=1, d=2): pass
f(1, 2, 3, 4)
f(1, 2)
f(a=1, b=2)
```
In addition this removes one inc/decref per arg because the function on the python value stack are already owned
so we can copy it into the frameobject directly.

In the future we can add support for more code object types.

`JIT_CALLSITES=1` enables this JIT (default: on)
`JIT_CALLSITES_TRIVIAL=1` JIT calls with no default args and no kwargs (default: off - normal path has nearly same speed)

The JIT is called from the AOT Profile functions. I tried before to profile via opcache mechanism in the interpreter and
emit the callsite jit as an IC directly in the calling function but this had a serious issue:
The function object in the cache is only borrowed so when JITing it could have been already destroyed and we crashed.

make measure perf is same/only slightly improved :(.
on this benchmark https://github.com/colesbury/nogil/blob/fb6aabede5f7f1936a21c2f48ec7fcc0848d74bf/benchmarks/call_benchmark.py one can see improvements:
```
                   no jit   with jit
call2              21.5 ns  19.9 ns*
call4              24.4 ns  21.2 ns*
call4_dflt         25.3 ns  20.7 ns
call4_kwd          33.0 ns  20.9 ns
call4_kwd_mismatch 32.7 ns  20.9 ns

No jit currently maybe something for the future:
call4_vararg_stararg    62.1 ns
call_vararg_stararg     34.3 ns
call_vararg4_stararg   112.3 ns
call_vararg4_kwd        77.5 ns
*with JIT_CALLSITES_TRIVIAL=1 enabled - the small speedup is not worth the additional code size -> disabled per default.
```
I think the remaining time would be much lowered by lazily creating `PyFrameObject` like cpython11 does.

Code size is only slightly increasing:
```
djangocms
jit: compiled 949 functions and 116 callsites
jit: 3255536 bytes used (88.7% of allocated)
jit: 26448 bytes are used by callsites
jit: compiled 436 functions and 88 callsites
jit: 1900112 bytes used (90.6% of allocated)
jit: 18864 bytes are used by callsites

flaskblogging
jit: compiled 386 functions and 77 callsites
jit: 1671680 bytes used (91.1% of allocated)
jit: 16592 bytes are used by callsites
```

The callsite JIT unfortunately does not have an impact on large benchmarks so the added complexity may speak against adding it but I think we can improve on it by adding support for more stuff and also by speeding up other overheads of calls this will hopefully show up more significant in the future. What do you think?

### DIFF
--- a/Include/funcobject.h
+++ b/Include/funcobject.h
@@ -33,6 +33,14 @@ typedef struct {
     PyObject *func_annotations; /* Annotations, a dict or NULL */
     PyObject *func_qualname;    /* The qualified name */
     vectorcallfunc vectorcall;
+    /* Version number for use by specializer.
+     * Can set to non-zero when we want to specialize.
+     * Will be set to zero if any of these change:
+     *     defaults
+     *     kwdefaults (only if the object changes, not the contents of the dict)
+     *     code
+     *     annotations */
+    uint32_t func_version;
 
     /* Invariant:
      *     func_closure contains the bindings for func_code->co_freevars, so
@@ -71,6 +79,8 @@ PyAPI_FUNC(PyObject *) _PyFunction_Vectorcall(
     PyObject *const *stack,
     size_t nargsf,
     PyObject *kwnames);
+
+uint32_t _PyFunction_GetVersionForCurrentState(PyFunctionObject *func);
 #endif
 
 /* Macros for direct access to these values. Type checks are *not*
--- a/Lib/test/test_sys.py
+++ b/Lib/test/test_sys.py
@@ -1213,7 +1213,7 @@ class SizeofTest(unittest.TestCase):
         check(x, vsize('5P2c4P3ic' + CO_MAXBLOCKS*'3i' + 'P' + extras*'P'))
         # function
         def func(): pass
-        check(func, size('13P'))
+        check(func, size('13Pi'))
         class c():
             @staticmethod
             def foo():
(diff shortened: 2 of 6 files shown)


kmod:
Oh this looks really cool!

I'm a bit mixed about it though, I wonder if we'll want to make broader changes to function calling and if we should do that all together.

One thing I don't quite understand is how the Faster CPython "noframe" patch speeds things up, since the frame allocation is pretty fast using the zombieframe, and I assume it has to do all the same work of local variable initialization / etc.

I guess I'm bringing that up because I suspect that we'll have to make some sort of larger change beyond just jitting/caching things but I still don't know exactly what it would be.

Anyway I'm fine with either merging it or not.

kmod:
Also because I wonder if the work is really in argument matching vs something else

PR #120, opened 2021-10-26
use bolt instrumentation and switch conda build to release

undingen:
This is based on https://github.com/pyston/pyston/pull/95. 
- verified that the speed is the same as the 2.3.1 release which used `perf lbr records`.
- had to include a not yet upstream merged bolt patch to support longer file paths.
- switches the conda build to use the release build
- had to add a lock to `pyston/run_profile_task.py` otherwise sometimes running the bolt pgo task failed
- switch skip funcs to use `_PyEval_EvalFrame_AOT_Interpreter.*` instead of `_PyEval_EvalFrame_AOT_Interpreter/1` because I fear that the compiler may not always call it exactly like this


kmod:
Looks great! Do you think it's worth just waiting for the bolt patch to be accepted to merge this?

undingen:
they merged my small bolt patch -> I updated the bolt submodule to the latest offical version

PR #122, opened 2021-11-02
 CI: run pyston/conda/build_pkgs.sh

undingen:
- now there is no difference between CI build and our local build_pkgs.sh
- don't build compiler-rt and bolt if available in pyston/label/dev channel
- renamed `opt` to `release` build
- uploading generated packages as github artifacts (makes it easy do download them - per default they are stored for 90days)

Not sure if I should also add the workaround for gcc-11:
```
build_base.sh:
# gcc-11 defaults to dwarf-5 which breaks bolt (Unsupported DWARFLocationEntry Kind)
CFLAGS=${CFLAGS}" -g -gdwarf-4"
```

### DIFF
--- a/.github/workflows/github-actions-ci.yml
+++ b/.github/workflows/github-actions-ci.yml
@@ -8,29 +8,21 @@ jobs:
         include:
           - build: unopt
             PYSTON_UNOPT_BUILD: 1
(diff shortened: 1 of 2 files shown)


kmod:
:+1:

PR #124, opened 2021-11-03
conda: prevent including system paths

undingen:
The github CI previously failed because it included `/usr/include` (I hacked around it by renaming `/usr/include/`).
Main problem was that we didn't enable cross-compilation (conda cpython rules do).
An additional problem was that if `setup.py` finds `dpkg-architecture` it will use it and include system include paths.
In the docker image used by `build_pkgs.sh` this program is installed but the
github CI container provides it.

- copy flags from the anaconda cpython recipe to enable cpython cross-compiling mode
- don't run `dpkg-architecture` if `CONDA_BUILD` is enabled
- add a check that we don't accidentally start including system headers again
- two tests start to fail if `_PYTHON_HOST_PLATFORM` is defined and must be disabled


kmod:
lgtm!

Are these all changes from the anaconda/conda-forge feedstock? I'm always wary when it feels like we're doing something that they're not doing.

undingen:
- The `build_base.sh` changes (setting `_PYTHON_HOST_PLATFORM` and the additional configure args) are coming from the anaconda/conda-forge cpython recipe.
- The failing tests skip changes are from me - I'm sure they would run in the same issues but they don't run the cpython tests currently (only mention they should do it in the future)
- Similar for the `dpkg-architecture` change one only hits this problem if the machine has it installed and the normal miniconda image does not have it installed. We run into it because the github CI had it installed.

I'm going right now over the additional things the cpython recipe does and will copy them if applicable for pyston and create a new PR.

kmod:
Ah ok cool, thanks!

PR #125, opened 2021-11-10
conda: add rules of cpython recipe 

undingen:
This adds a lot of code from the cpython conda recipe.
- use conda `CFLAGS` except `-fno-plt` which triggers a misscompile when running with bolt and `-pipe` (which has a fix which bolt has not yet merged)
- strip debug data from *.a
- move *.a to the libpython-static package
- don't package the tests
- patch sysconfig
- generate *.pyc files

Reduces release package artifact size to about 79MB from 152MB.

- I added some code which is not strictly necessary (e.g. code for non linux platforms) but which makes it easier to compare with the original cpython recipe.

- I did not add the the code where they copy in some preexisting sysconfigs: https://github.com/AnacondaRecipes/python-feedstock/blob/f91dedf533a5217d77dc48700e779d77820b68c7/recipe/build_base.sh#L427
I'm not sure why they need it / I guess it's for crosscompiling which we don't support. If we decide to also include we will need to patch them because they contain many cpython specific things.

### DIFF
--- a/pyston/conda/bolt/meta.yaml
+++ b/pyston/conda/bolt/meta.yaml
@@ -1,5 +1,5 @@
 {% set build_num = 0 %}
-{% set version = "2021.10.27" %}
+{% set version = "2021.11.13" %}
 
 package:
   name: bolt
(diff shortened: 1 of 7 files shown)


undingen:
My small patch to BOLT got merged -> I uploaded a new bolt package to the pyston channel and updated this PR to use it and removed the workaround for `CFLAGS=-pipe

kmod:
Looks great!

PR #126, opened 2021-11-15
POC: NaN boxing

undingen:
This weekend I was interested in writting a quick prototype for NaN boxing in our JIT:
 
This patch is a prototype for using NaN Boxing to store unboxed int32 and doubles.
In it's current form it's not safe because many cases where we need boxed values are not yet handled (also no deopts/exceptions etc..).
I just wanted to get a understanding of how complicated it would be and if this approach gives even any speedups.

This uses the same nan boxing as WebKits JS engine with JSVALUE64:
https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/runtime/JSCJSValue.h
We can store in a 64bit value the following: (the highest 15 bit mark what type it is)
 - 48bit pointer  (highest bits: 0000)
 - 32bit int      (highest bits: FFFE)
 - double         (highest bits: everything else)

Because for pointers the highest bits a zero a nan boxed pointer looks same as unboxed one.
In order to make this work we need to add a fixed constant value when storing a double 1<<49.

It will currently only use NaN boxing if the function name starts with `nan_`

**Results:**

`nan.py` example show what speedups can be reached in the best case:
```
- small ints: reduction in execution time by about 1/3
- floats    : reduction in execution time by about 1/2
- large ints: reduction in execution time by about 2/3
```
`nan_nbody.py` takes about twice es long with the current patch.
This is mostly because of the frequent box and unboxing of values.

So the results are quite mixed and I kind of doubt that we would see significant speedups in mixed workloads where only a few calcualtions can be unboxed.
For longs a much easier fix could be to add a freelist / faster allocator.


kmod:
This is really cool!

PR #127, opened 2021-11-15
conda: create a linux installer for our pyston packages

undingen:
only installs minimum required packages like miniconda/miniforge
- currently pointing to kmods and my channel because we don't have the packages yet the pyston channel

Install can be created by running:
$ conda install constructor -c conda-forge --override-channels
$ constructor --output-dir release/conda_pkgs pyston/conda/installer

I tested with fedora and could successful install (and import) tensorflow :)

I names it PystonConda not sure about the name... the default install location is all lowercase 'pystonconda'


kmod:
Awesome!

PR #128, opened 2021-11-17
Get many of the conda packages building

kmod:
- build_all.py is the main entrypoint, it figures out which packages have already been built, which ones need to be built, the order to do them, then builds them
- make_order.py does the work of figuring out the dependencies of packages and assembling them into a topological ordering
- make_config.py generates a variants file which is how we specify the python version to use
- a number of feedstocks needed patches, which happen via build_feedstock.sh
- I included almost all of the cpython patches that conda-forge applies


undingen:
Unfortunately the patches did not apply for me. I fixed it in: https://github.com/undingen/pyston_v2/commit/3567bdfe8be35926568882b439edeae935f281f4
But now I get a few test failures in `test_distutils` :/
```
======================================================================
ERROR: test_customize_compiler (distutils.tests.test_sysconfig.SysconfigTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/conda/conda-bld/pyston2.3-split_1637157360522/work/Lib/distutils/tests/test_sysconfig.py", line 124, in test_customize_compiler
    comp = self.customize_compiler()
  File "/opt/conda/conda-bld/pyston2.3-split_1637157360522/work/Lib/distutils/tests/test_sysconfig.py", line 104, in customize_compiler
    sysconfig.customize_compiler(comp)
  File "/opt/conda/conda-bld/pyston2.3-split_1637157360522/work/Lib/distutils/sysconfig.py", line 254, in customize_compiler
    if 'RANLIB' in os.environ and 'ranlib' in compiler.executables:
AttributeError: 'compiler' object has no attribute 'executables'

======================================================================
FAIL: test_get_msvcr (distutils.tests.test_cygwinccompiler.CygwinCCompilerTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/conda/conda-bld/pyston2.3-split_1637157360522/work/Lib/distutils/tests/test_cygwinccompiler.py", line 148, in test_get_msvcr
    self.assertRaises(ValueError, get_msvcr)
AssertionError: ValueError not raised by get_msvcr

======================================================================
FAIL: test_runtime_libdir_option (distutils.tests.test_unixccompiler.UnixCCompilerTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/conda/conda-bld/pyston2.3-split_1637157360522/work/Lib/distutils/tests/test_unixccompiler.py", line 83, in test_runtime_libdir_option
    self.assertEqual(self.cc.rpath_foo(), '-Wl,--enable-new-dtags,-R/foo')
AssertionError: '-Wl,-R/foo' != '-Wl,--enable-new-dtags,-R/foo'
- -Wl,-R/foo
+ -Wl,--enable-new-dtags,-R/foo



```

kmod:
Ok this should be good to review! The issues were
- The patches are to files with windows-style endings, and git was converting the patch files to unix-style endings so they were failing to apply
- A number of patches cause failures for us. I don't understand how they can work for conda-forge, I think this is something we should investigate but doesn't block this PR

undingen (approved):
This is great! I modified to upload the packages to my undingen repo and it seems to work fine.
Also thanks for figuring out the line ending thing I would not have realized that this is why the patches fail to apply.

undingen:
I just noticed that I think you forgot to `git add pyston/conda/patches/numba.patch`

PR #130, opened 2021-11-23
conda: supply patched sysconfigdata

undingen:
I forgot to port this code from the cpython recipe because I wrongly assumed it's only used when cross-compiling.
But it's actually necessary e.g. when trying to run pyperformance inside a conda pyston env it failed
because it could not find the 'x86_64-conda-linux-gnu-gcc' compiler (only available when explicitly installing the gcc package in
the conda env) and it should just use 'gcc' instead.

- merged the pyston changes into the cpython conda sysconfig
- pyperformance runs now
- not sure why the they have the 'replace-word-pairs.py'/'_FLAGS_REPLACE' code in the recipe when they are using the bundled sysconfig afterwards.
- I noticed issues with the static library:
-- I forgot to create the symlink in 'lib/python${VER}/config-${VER}-x86_64-linux-gnu/'
-- We use fat lto objects but if the compiler is not the exact same version as we used for compiling pyston you get an error.
   Thats why I added code to strip the LTO part (I noticed fedora is using similar code)
-- Unfortunately static compiling is still making problems because there is a mismatch with fpic:
   `relocation R_X86_64_32 against symbol 'PyBool_Type' can not be used when making a PIE object; recompile with -fPIE`


kmod:
It's a bit scary to me to have a completely different sysconfigdata for conda from the one that we otherwise use; do we know what's different between the two?

undingen:
yes I also don't like they approach they took. 
I just noticed that in the new 3.10 package conda forge replaces it with just running `sed` a few times on the existing sysconfig. I'm going to replace it with this approach because it looks much safer / lower maintenance.

PR #131, opened 2021-11-23
 conda: supply conda specific sysconfigdata

undingen:
This replaces #130 which a better approach which does not use a hardcoded sysconfig data file (as the cpython 3.8 recipe uses but instead uses the approach from 3.9/3.10)

I forgot to port this code from the cpython recipe because I wrongly assumed it's only used when cross-compiling.
But it's actually necessary e.g. when trying to run pyperformance inside a conda pyston env it failed
because it could not find the 'x86_64-conda-linux-gnu-gcc' compiler (only available when explicitly installing the gcc package in
the conda env) and it should just use 'gcc' instead.

- used the code from 3.9/3.10 conda recipe which does not use hard coded sysconfigdata
- requires the _CONDA_PYTHON_SYSCONFIGDATA_NAME patch
- pyperformance runs now
- I noticed issues with the static library:
-- I forgot to create the symlink in 'lib/python${VER}/config-${VER}-x86_64-linux-gnu/'
-- We use fat lto objects but if the compiler is not the exact same version as we used for compiling pyston you get an error.
   Thats why I added code to strip the LTO part (I noticed fedora is using similar code)
-- Unfortunately static compiling is still making problems because there is a mismatch with fpic:
   `relocation R_X86_64_32 against symbol `PyBool_Type' can not be used when making a PIE object; recompile with -fPIE`

### DIFF
--- a/Include/ceval.h
+++ b/Include/ceval.h
@@ -97,7 +97,7 @@ PyAPI_FUNC(int) _Py_CheckRecursiveCall(const char *where);
 */
 PyAPI_DATA(int) _Py_CheckRecursionLimit;
 
-static inline void* _stack_pointer() {
+static inline void* _stack_pointer(void) {
     void* sp;
     __asm( "mov %%rsp, %0" : "=rm" ( sp ));
     return sp;
--- a/Include/tracemalloc.h
+++ b/Include/tracemalloc.h
@@ -38,7 +38,7 @@ PyAPI_FUNC(PyObject*) _PyTraceMalloc_GetTraceback(
 
 #else
 
-static inline int _tracemalloc_success() {
+static inline int _tracemalloc_success(void) {
     return -2;
 }
 
--- a/pyston/conda/pyston/build_base.sh
+++ b/pyston/conda/pyston/build_base.sh
@@ -11,7 +11,12 @@ rm -rf build
 CFLAGS=$(echo "${CFLAGS}" | sed "s/-fno-plt//g")
 
 _OPTIMIZED=yes
+VERFULL=${PKG_VERSION}
 VER=${PYTHON_VERSION2}-pyston${PYSTON_VERSION2}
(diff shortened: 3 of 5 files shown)


undingen:
With this PR applied the default compiler changes depending on if `_CONDA_PYTHON_SYSCONFIGDATA_NAME` is set or not. (and it seems to get automatically set when a conda compiler package is included in a recipe). This way the default system compiler gets used if one just pip installs something in a conda environment without a installed compiler (which did not work before).
```
(pyston4) undingen@pcw:~/dev/pyston_v2$ pyston -msysconfig | grep "CC ="
	CC = "gcc -pthread -B /home/undingen/miniconda3/envs/pyston4/compiler_compat"
(pyston4) undingen@pcw:~/dev/pyston_v2$ _CONDA_PYTHON_SYSCONFIGDATA_NAME="_sysconfigdata_x86_64_conda_cos6_linux_gnu" pyston -msysconfig | grep "CC ="
	CC = "x86_64-conda_cos6-linux-gnu-gcc -pthread"
```

kmod:
Maybe I'm being dense, but how does setting that variable affect the default compiler? Because it picks up a different sysconfig file?

I think the patch you included is going to cause the build to fail, because it renames the sysconfigdata file which our Makefile needs to know about

kmod:
ok, lgtm!

undingen:
I rebased the PR and incremented the pyston2.3 package build number to 6.
Everything seems to still work when installing this package and using the 'pyston' channel packages so I think this should be ready for merging but just want to make sure you are okay too now that we already have build packages. I think this is a slight improvement because it improves compatibility if someone is pip installing packages inside a conda environment. Most prominently without it pyperformance does not work per default so quick performance comparison is harder.

kmod:
oh dang I thought we already merged this haha

PR #132, opened 2021-11-24
conda: fix 'vim' package

undingen:
apply a patch which changes the library and include path to pystons

Also sneaked in a few small changes to `build_all.py`


kmod:
Thanks for the fixes!

I'm not thrilled about hardcoding the pyston version in the vim package like that, but at least it should complain loudly next time we bump the version number.

PR #133, opened 2021-11-25
last few conda fixes

kmod:
- add the numba patch I forgot to git add
- run_constrain pyston to have a version of cython that works with it

I verified that the cython constraint will correctly prevent us from installing a version of cython without our patch, but since there's currently no working version of it anything depending on cython will fail to build. once we get a pyston version made I'll update the constraint to point to it. I think it's worth checking this in for now so that I can build all the non-cython-dependent packages and make sure the cython ones fail


undingen (approved):
LGTM

PR #134, opened 2021-11-25
conda: fix memory corruption in cython 0.29

undingen:
unfortunately our fix #4200 is only in the 0.30a releases
and not in the 0.29 branch.
This backports the fix.


kmod:
Thanks for doing this!

PR #137, opened 2021-11-30
conda: fix torchvision package and make installer ready for release

undingen:
Not sure what version number it should be thought maybe 1.0 is good
but it does bundle pyston 2.3.1 so maybe it should be 2.3.1 instead?

### DIFF
--- a/pyston/conda/build_feedstock.sh
+++ b/pyston/conda/build_feedstock.sh
@@ -210,6 +210,19 @@ if [ "$PACKAGE" == "vim" ]; then
     sed -i "/source:/a \  patches:\n    - pyston.patch" recipe/meta.yaml
 fi
 
+if [ "$PACKAGE" == "torchvision" ]; then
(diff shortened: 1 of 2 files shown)


pyston/conda/build_feedstock.sh
```diff
@@ -210,6 +210,20 @@ if [ "$PACKAGE" == "vim" ]; then
     sed -i "/source:/a \  patches:\n    - pyston.patch" recipe/meta.yaml
 fi
 
+if [ "$PACKAGE" == "torchvision" ]; then
+    cp $THISDIR/patches/vim.patch recipe/pyston.patch
```
kmod: I think this line is spurious

kmod (approved):
lgtm!

pyston/conda/build_feedstock.sh
undingen: good catch!

PR #138, opened 2021-12-01
Add a Dockerfile with a pre-installed conda environment

kmod:
and push it to docker hub (which I've done already)

you can try it out with
```
docker run -it pyston/conda bash
```


undingen (approved):
LGTM

PR #139, opened 2021-12-02
A few more conda packaging updates

kmod:
mostly related to uploading the packages to the "real" channel, but also get a few more libraries working.


undingen:
LGTM

PR #140, opened 2021-12-02
Build multiple versions of the same package

kmod:
This is based on #139, so only the commits starting with "Initial stab at building multiple versions" are part of this PR

I noticed that a few of the next 500 packages are failing because they are pinned to old versions of libraries. We tracked the dependency correctly and uploaded the latest version of their dependency, but they're pinned to an old one and will still fail.

This PR deals with one part of this, which is making it easy to build multiple versions of the same package. Part of that is no longer having to manually search the feedstock git log to find a revision corresponding to the package version we want.

I think this will be increasingly common since our users will possibly be pinned to old versions of libraries and we'll want to generate those as well.


undingen:
Very nice improvement!

PR #141, opened 2021-12-03
CI: upload core dumps

undingen:
If the build crashes we will now generate a core dump and upload
it together with the conda build directory.

I tested it with a small artificial segfault and it worked fine: https://github.com/undingen/pyston_v2/actions/runs/1537123679


kmod:
awesome!

PR #142, opened 2021-12-07
Get a few more packages working

kmod:
I also set it up to start building multiple versions of packages like numpy, but the build is still ongoing.


undingen:
LGTM, I could confirm that upgrading the `pyston2.3-2.3.1-7_23_pyston` (package containing this PR) to a future `*-8*` build (only changed the build number left everything else the same) worked fine. All site-packages survived. But upgrading from the previous `*-6*` package fails but I think this is not a problem because outside of internal testing nobody should be using it currently.

PR #145, opened 2021-12-10
Conda: upgrade to Cython 0.29.25 and fix 'sed' cmds in feedstock script

undingen:
- Cython 0.29.25 contains our fix so we don't need to apply our patch anymore.
- Had to update pyston2.3 build number too because the runtime constrained changed (think this is a bit annoying but afterwards every release should just work / alternative is to always append `_pyston` to the build string).
- Figured out that many packages currently don't build if the `sed` script contains `\n`.

I build a few test packages on my undingen channel and also tried upgrading from the offical `*_7*` pyston package to the newly `*_8*` and it seemed to work fine and it also installed the new cython (but one needs to upload the pyston package before building cython because else the build fails because of the `cython >=0.29 *_pyston` constrain).

### DIFF
--- a/pyston/conda/build_feedstock.py
+++ b/pyston/conda/build_feedstock.py
@@ -102,9 +102,15 @@ def getRecipeSedCommands(feedstock, version):
         return ("s/  sha256: {{ sha256 }}/  sha256: {{ sha256 }}\n  patches:\n    - pyston.patch/",)
 
     if feedstock == "cython":
+        if version == "latest":
+            return ()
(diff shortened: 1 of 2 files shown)


kmod:
Looks great, thanks!

kmod:
I'll build the new packages

kmod:
Ok I built and uploaded the new packages!

Previously I had been deleting old packages once we had a better one up, but now that we've announced them I think it's time to leave them there.

undingen:
Thanks sounds good!

PR #146, opened 2021-12-13
jit: optimize 'is'/'is not' cmps, PyObject_IsTrue and UNARY_NOT 

undingen:
Small optimizations which improve the performance for the following things:
- generate AOT trace for `PyObject_IsTrue`:
     used by the normal compare and jump opcodes but also now by `UNARY_NOT`
- directly emit `UNARY_NOT` and make it use the `PyObject_IsTrue` trace
- directly emit `is` and `is not` comparisons: this saves a function call and error check

Seeing about 15% speedup on nanobenchmark and very slight speedup on `make measure`.

-----
The traces for `PyObject_IsTrue` can be improved if we tell pystol that
```C
// trace of PyObject_IsTrueLong1
if (unlikely(!(o0->ob_type == &PyLong_Type))) {
   [...]
   return ret;
}
// code inlined from PyObject_IsTrue(PyObject *v) which should be removed:
if (o0 == Py_True)  // remove we know Py_TYPE(Py_True) == PyBool_Type so they can't be equal
    return 1;
if (o0 == Py_False) // remove
    return 0;
if (o0 == Py_None) // remove 
    return 0;
[...] // real code following
```
I'm not sure how to add this to the facts system :/. While not high priority I thought you may be interested in adding it because it's seems like a fun optimization which may trigger for quite a few things.

### DIFF
--- a/Include/internal/aot_ceval_jit_helper.h
+++ b/Include/internal/aot_ceval_jit_helper.h
@@ -22,7 +22,6 @@ extern "C" {
 #define JIT_HELPER_WITH_NAME_OPCACHE_AOT2(name_, py1, py2) PyObject* JIT_HELPER_##name_(PyObject* name, PyObject* py1, PyObject* py2, _PyOpcache *co_opcache)
 
 
-JIT_HELPER1(UNARY_NOT, value);
 JIT_HELPER1(PRINT_EXPR, value);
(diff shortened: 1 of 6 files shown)


kmod:
awesome good idea!

I'll look into why the trace optimizer didn't catch those cases, I thought I had added that already

PR #148, opened 2021-12-15
Improve PyObject_IsTrue traces

kmod:
Remove most of the comparisons with True/False/None
Teach it that all objects of type PyNone_Type are the None object


undingen (approved):
LGTM, thanks for implementing this would have taken me a long time :sparkles:

PR #150, opened 2021-12-16
fix memory leak in PyLong_FromString

### LINKED ISSUE #149: Memory leak when using pyexcel.iget_records
When using pyexcel to stream CSV files with Pyston, I get a memory leak that does not occur when running the same code with Python 3.8.

I created a minimal example that creates a random CSV with ~3M lines and reads it using [`pyexcel.iget_records`](http://docs.pyexcel.org/en/latest/generated/pyexcel.iget_records.html). Repo with steps to reproduce: https://github.com/PLPeeters/pyston-pyexcel-leak.

When monitoring `docker stats`, the memory increases as follows from start to finish:

* Python 3.8: 32 MB -> 33 MB
* Pyston 2.3.1: 39 MB -> 723 MB

When changing the CSV delimiter to `;`, which reduces the number of fields per line from 3 to 1, memory before exiting is reduced to 300 MB, so the number of fields pyexcel has to process seems to play a role. I'm guessing there's some kind of mechanism pyexcel relies on in one way or another that is broken in Pyston and causes it to leak memory (@chfw could have some insights?).

As a side note, in my quest to get my example down to the bare minimum, I checked if using `csv.reader` would reproduce this bug as well; it doesn't.

undingen:
We forgot to decref a value in the error path of this function.
It's easily reproducible by:
```python
for i in range(1000000):
    try:
        int("ca4b9344", 10)
    except:
        pass
```
This fixes #149

### DIFF
--- a/Objects/longobject.c
+++ b/Objects/longobject.c
@@ -2651,15 +2651,20 @@ digit beyond the first.
         goto done;
 
     PyObject* strobj_repr = PyObject_Repr(strobj);
-    if (!strobj_repr)
+    if (!strobj_repr) {
+        Py_DECREF(first);
         goto done;
+    }
 
     PyObject* strobj_repr_truncated = PyUnicode_Substring(strobj_repr, 0, 200);
     Py_DECREF(strobj_repr);
-    if (!strobj_repr_truncated)
+    if (!strobj_repr_truncated) {
+        Py_DECREF(first);
         goto done;
+    }
 
     PyObject* err = PyUnicode_Concat(first, strobj_repr_truncated);
+    Py_DECREF(first);
     Py_DECREF(strobj_repr_truncated);
 
     PyErr_SetObject(PyExc_ValueError, err);


kmod:
oh dang thanks for catching this

PLPeeters:
Damn, that was an insanely quick fix, thanks a bunch! Any chance of getting a new Docker image with this fix out soon? 😇

PLPeeters:
@kmod Any chance we could get a 2.3.2 release with this included?

undingen:
@PLPeeters sorry for not responding. We have a found a few more bugs - most importantly we also have to fix the incompatibility with setuptools v60 (not sure if this even ends up requires a change to pyston). I would like to wait until we fixed the problems and afterwards release v2.3.2. Can you wait a few more days?

PLPeeters:
@undingen No worries, that makes sense; I can wait a few more days!

kmod:
Hi @PLPeeters we just put up [2.3.2](https://github.com/pyston/pyston/releases/tag/pyston_2.3.2), sorry to keep you waiting!

PLPeeters:
@kmod Awesome, thanks for the update! Any ETA for the corresponding Docker images?

kmod:
Oops sorry, forgot to do that part. They're up now!

PR #154, opened 2022-01-10
AOT: fix crash inside isinstance trace

undingen:
The problem is that right now a trace can't call another trace safely because the called trace can use `SET_JIT_AOT_FUNC` which will look at the return address to find the address to patch.
Normally this points to JITed code but in this case it will be another AOT func.
The frame looks something like this:
`JIT func -> call_functionisinstanceLong3 -> call_functionisinstance3 -> tries to modify return address...`

Clang 13 exposes `__attribute__((musttail))` which should fix the issue unfortunately it got just released.
But there is luckily a simple workaround: change the pointer to the AOT trace but call the unspecialized case.
The JITed code will call the AOT trace next time so we only end-up calling one more time the unspecialized case than strictly necessary.

The issue was really hard to trigger because it only happened when the isinstance function changed but not if only the args changed.

### DIFF
--- a/pyston/aot/aot_gen.py
+++ b/pyston/aot/aot_gen.py
@@ -240,6 +240,24 @@ class Handler(object):
         else:
             return self.case.unspecialized_name
 
+    def _emitDeopt(self, deopt_to, pass_args, f):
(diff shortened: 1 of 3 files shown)


kmod:
:+1:

PR #155, opened 2022-01-11
OSR: track all backward jumps

### LINKED ISSUE #153: OSR doesn't trigger if a loop ends with a conditional
@undingen you implemented something that will tier up from inside a for loop, but I just ran into a case where it doesn't happen:

```python
def f():
    for i in range(100000000):
        if i == -1:
            print(i)

import dis
dis.dis(f)

f()
```

Here's the disassembly:
```
  2           0 LOAD_GLOBAL              0 (range)
              2 LOAD_CONST               1 (100000000)
              4 CALL_FUNCTION            1
              6 GET_ITER
        >>    8 FOR_ITER                20 (to 30)
             10 STORE_FAST               0 (i)

  3          12 LOAD_FAST                0 (i)
             14 LOAD_CONST               2 (-1)
             16 COMPARE_OP               2 (==)
             18 POP_JUMP_IF_FALSE        8

  4          20 LOAD_GLOBAL              1 (print)
             22 LOAD_FAST                0 (i)
             24 CALL_FUNCTION            1
             26 POP_TOP
             28 JUMP_ABSOLUTE            8
        >>   30 LOAD_CONST               0 (None)
             32 RETURN_VALUE
```

The problem is that this loop takes POP_JUMP_IF_FALSE as the backedge, but only JUMP_ABSOLUTE is coded to trigger OSR. (I assume there must have been a peephole step where it noticed that the POP_JUMP_IF_FALSE target was a JUMP_ABSOLUTE, and it fused the two jumps together.)

undingen:
before we missed a few cases because we only handled JUMP_ABSOLUTE.

The percentage of compiled functions increased by about 2-3% on flaskblogging which looks fine to me.
Make measure stays about the same. Have not tried if it speeds up microbenchmarks but definitely nice to have more predictable OSR behavior.

This fixes #153

### DIFF
--- a/Python/aot_ceval.c
+++ b/Python/aot_ceval.c
@@ -1404,6 +1404,17 @@ _PyEval_EvalFrame_AOT_Interpreter(PyFrameObject *f, int throwflag, PyThreadState
 #define JUMPTO(x)       (next_instr = first_instr + (x) / sizeof(_Py_CODEUNIT))
 #define JUMPBY(x)       (next_instr += (x) / sizeof(_Py_CODEUNIT))
 
+// JUMPTO which is counting backward jumps in order to todo OSR
+#define JUMPTO_WITH_OSR(x)  \
+    do { \
+        _Py_CODEUNIT* old_instr = next_instr; \
+        JUMPTO(x); \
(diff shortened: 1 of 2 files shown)


kmod:
awesome!

PR #157, opened 2022-01-18
JIT: fix crash in Is/Is not comparisons

undingen:
My newly introduced #146 optimization contained a bug which caused crashes.
Problem was that when both sides of the comparison are owned values and the first refcount reaches 0.
The call to `_Py_Dealloc` clobbers all registers including the one where we stored the second variable to decref.
This caused us to decref random memory :/.


kmod:
ah good catch, this seems annoying to discover

PR #158, opened 2022-01-18
CI: run external testsuites

undingen:
This significantly enhances our CI tests. And should make it much less likely for regressions to sneak in.

- add new non conda build job which uses regular unopt non-conda build but with ubuntus llvm libs
     - needs to run inside own docker image so we can control exactly what dependencies get installed
     - could not use the conda job because of dependency issues while installing the dependencies for the external tests.
     - we are building own ubuntu 20.04 based image because some tests fail when run as root and I found this the easiest and cleanest solution
- moved `test_numpy` to the `EXTERNAL_TESTSUITES`
- only run the external testsuites with default JIT flags to reduce build time
- `test_numpy` failed for me. It seemed the `usercustomize.py` script did not get run so I switched to putting in a `.pth` file

Runtime seems to be about 2h15min - will update when all CI jobs have finished

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -1,0 +1,31 @@
+FROM ubuntu:latest
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# set up locale
+RUN apt-get update
+RUN apt-get -y install locales locales-all
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+# copied from https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user
+ARG USERNAME=nonroot
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+# Create the user
+RUN groupadd --gid $USER_GID $USERNAME \
+    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
+    #
+    # [Optional] Add sudo support. Omit if you don't need to install software after connecting.
+    && apt-get update \
+    && apt-get install -y sudo \
+    && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME \
+    && chmod 0440 /etc/sudoers.d/$USERNAME
+
+# ********************************************************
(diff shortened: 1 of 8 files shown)


undingen:
Ok so this is super strange without the second commit django is always failing:

>test: change order of cpython and pyston run
>
>Why does cpython when run before pyston pickup more (15035) tests
>than pyston (14458) on test_django. But when the order is swapped both only pick up 14458 and the test succeeds.
>
>It seems to not run the tests in the 'test/cache/' directory.

I can't understand what the difference is - maybe you know?
It also seems all the tests got picked up in the commits test run: 15035 not 14458 like it wrote in the commit message. (I also run it several times and they all succeeded - but only when the order of cpython and pyston are swapped)

kmod:
I've seen something that might be related -- somehow, sometimes a file (a test data file I think) gets deleted from the django directory, which changes something about the testing process which causes the comparison to fail. On my local machine I can notice this by running `git status` and it will report a deleted file.

I put together a quick test and started it running to see if this is what's happening:
https://github.com/kmod/pyston/actions

If this isn't it I'm not sure what's up :/

undingen:
good point. I notices this once but than thought it must have been my editor or so (because it was some file with "~" in the name) playing games. But make sense that it was the actually the test run which modified it.

Based on your test runs it seems this is not the problem here?

kmod:
Yeah that doesn't seem to be it so I don't know what it is :/  Some random ideas
- can we reproduce this locally? if so are there any changes to the django/ directory after the test?
- does running it without parallelism change anything?

if both of those are "no" then I'm very confused

undingen:
- I could not reproduce it on my local machine but did manage to reproduce it now inside the docker image.
- running: `make test` instead of `make unopt -j$(nproc)` did not fix the CI
- it's faster to reproduce by changing to only `EXTERNAL_TESTSUITES:=django`

kmod:
Lgtm! Here's my fix:
https://github.com/kmod/pyston/commit/51a39f712cdf578ea7bc3e7bd180b19e327a9c4f

undingen:
thanks for tracking down the issue

PR #159, opened 2022-01-20
Upgrade to newest DynASM for better ARM64 support

undingen:
It contains vreg support for ARM64 + a few encoding fixes.

For amd64 `|.aword =>inst_idx` is now emitting a 8 byte value instead of 4.
Which makes sense but we took advantage of this so I'm now emitting the table manually in order
to not consume twice the memory.

`| mov Rd(r_dst), [Rq(r_inst_idx)*2 + ->opcode_addr_begin]` is now broken.
It will emit a 8byte `->opcode_addr_begin` instead of the 4byte immediate value required by the instruction set.
I filled a bug report https://github.com/LuaJIT/LuaJIT/issues/801 and added a workaround for now.

### DIFF
--- a/Python/aot_ceval_jit.c
+++ b/Python/aot_ceval_jit.c
@@ -1007,17 +1007,18 @@ static void* get_aot_func_addr(Jit* Dst, int opcode, int oparg, int opcache_avai
     return get_addr_of_aot_func(opcode, oparg, opcache_available);
 }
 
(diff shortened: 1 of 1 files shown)


undingen:
I also run the our integration tests on another branch and the succeeded so should be good

kmod:
:+1:

PR #161, opened 2022-01-20
CALL_METHOD inline caches

kmod:
I was thinking about the second line of the following code snippet:
```
l = []
l.append(x)
```

Ideally we would emit code that looks something like
```
if (PyList_CheckExact(l))
  PyList_Append(l, x);
else
  slowpath();
```

However here's what we do currently:

```
LOAD_FAST:

LOAD_METHOD:
set lasti
test for tracing
jump if tracing
load l from frame
incref l
move l onto the value stack
adjust the value stack pointer
load the type
check the version tag
not-taken jump to guard failure
load the dict offset
test the dict offset
take jump to dict check success
load the constant attribute value into rax
incref it
put it onto the stack
copy l from rdi to rax

CALL_METHOD:
set lasti
test for tracing
jump if tracing
copy rax (l) onto the value stack
adjust the value stack
copy tstate to rdi [arg1]
copy value stack pointer to rsi [arg2]
set rdx [arg3] to zero (using an xor)
cmpq + sbb (adjust arg3 by one if a method is found)
call aot function

call_functionMethodDescr2
[this is abbreviated, it's quite a few assembly instructions]
push 5 regs
guard on oparg
fetch f->ob_type
cmp it to PyMethodDescr_Type
jne if failed
check tstate->tracing
load the vectorcall
check if its null (for tpcall style)
load method_vectorcall_noargs
check if its that
call _PyObject_RealIsSubclass
check for recursion limit being exceeded
load func->d_method->ml_meth
test if its null
call it
decref something
decref something else
return

adjust vsp back by 16
test if rax is null (exception)
je if exception (not taken)

POP_TOP
mov rax to rdi
decref rdi
je if deallocated
```

The basis of this PR is that we have an extremely good guess of the exact object that will be called in the CALL_METHOD bytecode, since it is stored in the inline cache for the corresponding LOAD_METHOD bytecode.

This PR specifically targets calling "method descriptors" (methods that are builtin fuctions) of builtin classes.

So with this PR the above simplifies to

```
LOAD_FAST

LOAD_METHOD
set lasti
test for tracing
jump if tracing
load l
incref l
store l to the value stack
adjust value stack pointer
check if this is list
jump for guard failure
mov list.append to rax
store rax to the value stack
mov l back to rax

LOAD_FAST

CALL_METHOD
set lasti
test for tracing
jump for tracing
load x
incref x
move x to the value stack
move rax (l) to the value stack
adjust the value stack
load tstate->use_tracing
check if tstate->use_tracing is nonzero
jump if it's nonzero
load the method object from the value stack
check if its null
jump if its null
load self (l) from the value stack
check if its type is list
jump if its not a list
load x from the value stack into rsi (arg2)
call list_append
load x from the value stack
decref it
jump if dealloc
load al from the value stack
decref it
jump if dealloc
adjust the value stack
check for an exception
jump for exception

POP_TOP
move the result from rax to rdi
decref it
jump if deallocd        
```

Which includes the following optimizations:
- no more incref+decref of list.append (it's immortal now)
- Instead of checking the version tag and then the dictoffset, we just check if the type is list
- inline call_functionMethodDescr2
- skip all the guards inside call_functionMethodDescr2 once we know the exact value of the callable

This results in a pretty large reduction in the overhead of these calls. Here are some timings of these kinds of functions -- this is the total time to call the function, including all the bookkeeping overhead and the actual work of the function itself:

```
list.clear: 6.8ns before, 3.6ns now
list.append: 12ns before, 10ns now
list.insert: 20ns before, 15ns now
dict.pop: 8ns before, 6ns now
str.startswith: 17ns before, 14ns now
```

Unfortunately the behavior of this PR seems pretty different on my two machines, which I presume is due to one being an AMD cpu and the other being an Intel CPU. This PR is not a ton of help on the AMD cpu, and sometimes seems to hurt it (seems to be due to extra icache pressure from the additional code). On my intel cpu it speeds up the macrobenchmarks by about 1%:

```
  system 31.98+-0.21              net +0.0%
b704af68 25.39+-0.03              net -20.6% Add some more trace examples
44c2a121 25.32+-0.04  -0.3%+-0.2% net -20.8% Add a new loadattr cache type
8b2b8964 25.30+-0.02  -0.1%+-0.1% net -20.9% Implement CALL_METHOD "hinting"
5e89aee4 25.17+-0.03  -0.5%+-0.1% net -21.3% Specialize for 1-arg calls as well
b38929d1 25.20+-0.02  +0.1%+-0.1% net -21.2% Add fastcall support to the method hints
709c8bac 25.21+-0.03  +0.0%+-0.1% net -21.2% Small makefile bug
c6ef5680 25.18+-0.03  -0.1%+-0.1% net -21.3% Cap how many decrefs we'll inline
8a644ce9 25.13+-0.02  -0.2%+-0.1% net -21.4% Create specialized versions of decref_array
6fe7c029 25.24+-0.02  +0.4%+-0.1% net -21.1% Test that exposes a segfault
0ab3d9bd 25.12+-0.04  -0.5%+-0.1% net -21.5% Add an additional guard to fix this bug
7711c177 25.11+-0.03  -0.1%+-0.1% net -21.5% Make builtin methods immortal to avoid inc/decrefs
c4b1b2a9 25.19+-0.02  +0.3%+-0.1% net -21.2% Handle VARARGS methods
daeb0440 25.10+-0.04  -0.4%+-0.1% net -21.5% Support method_call ics for methods that take keyw
2b07f84b 25.08+-0.05  -0.1%+-0.2% net -21.6% Now that we handle all method types, clean this up
51105afe 25.11+-0.01  +0.1%+-0.2% net -21.5% Add in a check for tracing
```

I checked one of the benchmarks and all the remaining calls to call_functionMethodDescr* are due to calling method descriptors of subclasses of builtin classes (such as OrderedDict), which I think is possible to handle but seems to be a pretty minor case.

Future work:
- We could do this same optimization with LOAD_GLOBAL + CALL_FUNCTION pairs
- I'm not sure how to do this, but if we were able to fuse the LOAD_METHOD and CALL_METHOD bytecodes together, then we could remove the duplicate guards, and get rid of a bunch of stack manipulations and refcounting operations. Unfortunately those two bytecodes are not next to each other except for 0-arg calls.
- There's still a lot of value-stack manipulation in the inline cache which I think could be deferred to the slowpath
- We know that list_append always returns None (or NULL), so we could skip decreffing it.

### DIFF
--- a/Python/aot_ceval_jit.c
+++ b/Python/aot_ceval_jit.c
@@ -2315,7 +2315,7 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
                 CallMethodHint* hint = Dst->call_method_hints;
                 Dst->call_method_hints = hint->next;
 
-                if (hint->attr) {
+                if (hint->attr && hint->meth_found) {
                     int num_args = oparg + hint->meth_found; // number of arguments to the function, including a potential "self"
                     int num_vs_args = num_args + 1; // number of python values; one more than the number of arguments since it includes the callable
 
@@ -2324,11 +2324,22 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
                         void* funcptr = method->d_method->ml_meth;
 
                         if (funcptr && _PyObject_RealIsSubclass((PyObject*)hint->type, (PyObject *)PyDescr_TYPE(method))) {
+                            // Strategy:
+                            // First guard that meth != NULL.
+                            // We need this to verify that the object in the "self" stack slot
+                            // is actually the self object and not a non-method attribute.
+                            //
+                            // Then guard that self->ob_type == hint->type
+                            //
                             // We skip the recursion check since we know we did one when
                             // entering this python frame.
-                            if (method->vectorcall == method_vectorcall_NOARGS && num_args == 1) {
-                                | mov arg1, [vsp - 8 * num_args] // self
 
+                            if (method->vectorcall == method_vectorcall_NOARGS && num_args == 1) {
+                                | mov arg1, [vsp - 8 * num_vs_args] // callable
+                                | test arg1, arg1
+                                | je >1
+
+                                | mov arg1, [vsp - 8 * num_args] // self
                                 | cmp_imm_mem [arg1 + offsetof(PyObject, ob_type)], hint->type
                                 | jne >1
 
@@ -2336,8 +2347,11 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
 
                                 wrote_inline_cache = 1;
                             } else if (method->vectorcall == method_vectorcall_O && num_args == 2) {
-                                | mov arg1, [vsp - 8 * num_args] // self
+                                | mov arg1, [vsp - 8 * num_vs_args] // callable
+                                | test arg1, arg1
+                                | je >1
 
+                                | mov arg1, [vsp - 8 * num_args] // self
                                 | cmp_imm_mem [arg1 + offsetof(PyObject, ob_type)], hint->type
                                 | jne >1
 
@@ -2347,6 +2361,10 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
 
                                 wrote_inline_cache = 1;
                             } else if (method->vectorcall == method_vectorcall_FASTCALL) {
+                                | mov arg1, [vsp - 8 * num_vs_args] // callable
+                                | test arg1, arg1
+                                | je >1
+
                                 | mov arg1, [vsp - 8 * num_args] // self
 
                                 | cmp_imm_mem [arg1 + offsetof(PyObject, ob_type)], hint->type
--- a/Python/aot_ceval_jit.c
+++ b/Python/aot_ceval_jit.c
@@ -227,7 +227,9 @@ Py_ssize_t lookdict_split(PyDictObject *mp, PyObject *key, Py_hash_t hash, PyObj
 PyObject * method_vectorcall_NOARGS(PyObject *func, PyObject *const *args, size_t nargsf, PyObject *kwnames);
 PyObject * method_vectorcall_O(PyObject *func, PyObject *const *args, size_t nargsf, PyObject *kwnames);
 PyObject * method_vectorcall_FASTCALL(PyObject *func, PyObject *const *args, size_t nargsf, PyObject *kwnames);
+PyObject * method_vectorcall_FASTCALL_KEYWORDS(PyObject *func, PyObject *const *args, size_t nargsf, PyObject *kwnames);
 PyObject * method_vectorcall_VARARGS(PyObject *func, PyObject *const *args, size_t nargsf, PyObject *kwnames);
+PyObject * method_vectorcall_VARARGS_KEYWORDS(PyObject *func, PyObject *const *args, size_t nargsf, PyObject *kwnames);
 
 static void decref_array(PyObject** vec, int n) {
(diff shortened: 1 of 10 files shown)


Python/aot_ceval_jit.c
```diff
@@ -2246,6 +2309,133 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
...
+
+                            JIT_ASSERT(sizeof(tstate->use_tracing) == 4, "");
+                            | mov eax, [tstate + offsetof(PyThreadState, use_tracing)]
+                            | test eax, eax
+                            | jne >1
+
+                            | mov arg1, [vsp - 8 * num_vs_args] // callable
+                            | test arg1, arg1
+                            | je >1
+
```
undingen, line 2357: Replacing this mov+test compinations with the following code should save two instructions. I guess this will likely not change perf but I think it makes the code also a bit more readable because it's clear that it's just a guard and later code is not reading the register.
```
                            | cmp dword [tstate + offsetof(PyThreadState, use_tracing)], 0
                            | jne >1

                            | cmp dword [vsp - 8 * num_vs_args], 0 // callable
                            | je >1
```

Python/aot_ceval_jit.c
```diff
@@ -2246,6 +2309,133 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
...
+                if (hint->attr && hint->meth_found) {
+                    int num_args = oparg + hint->meth_found; // number of arguments to the function, including a potential "self"
+                    int num_vs_args = num_args + 1; // number of python values; one more than the number of arguments since it includes the callable
+
+                    if (hint->attr->ob_type == &PyMethodDescr_Type) {
+                        PyMethodDescrObject* method = (PyMethodDescrObject*)hint->attr;
+                        void* funcptr = method->d_method->ml_meth;
+
+                        if (funcptr && _PyObject_RealIsSubclass((PyObject*)hint->type, (PyObject *)PyDescr_TYPE(method))) {
+                            wrote_inline_cache = 1;
```
undingen, line 2334: if you already set `wrote_inline_cache=1` here the later assignments can be removed.

Python/aot_ceval_jit.c
```diff
@@ -3129,6 +3332,14 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
     free(Dst->known_defined);
     Dst->known_defined = NULL;
 
+    CallMethodHint *hint = Dst->call_method_hints;
+    Dst->call_method_hints = NULL;
+    while (hint) {
+        CallMethodHint *new_hint = hint->next;
```
undingen, line 3346: I think this code to free the memory never gets reached because AFAIK all the CALL_METHODs remove an element from the list. If this is here to catch unpaired LOAD_METHOD/CALL_METHOD we require a `free(hint)` at the end of the IC.

undingen (commented):
Wow this is a very cool PR.
As said on the call seeing the overhead written down like you did make it look quite bad :-P glad you managed to remove some of it.
I have three small comments:
- some wrote_inline_cache=1 can be removed
- generate two instructions less per method ic
- missing free(hint)

I could also confirm about 1% speedup on make measure on my intel alder lake

kmod:
Ah good catches! I pushed a couple new commits that address them.

Python/aot_ceval_jit.c
undingen: good think you also catched my `cmp dword error` which should have been a `qword` like you wrote.

PR #163, opened 2022-01-20
Change inc/dec instructions to adds

kmod:
I noticed when comparing assembly code of our jit vs the
aot traces that the jit had "dec" and the aot traces had "add -1".

Apparently add is faster than dec, purportedly due to not propagating
any flags (which would cause dependencies), and it looks like
both clang and gcc will use "add -1" instead of dec (except in -Os).

I wasn't able to measure a speedup but I think we should change it anyway.


undingen (approved):
LGTM

PR #164, opened 2022-01-21
Argument parsing: fast-path this check

kmod:
PyFloat_Check is somewhat expensive because it doesn't use the fast
bitmask method. And presumably this error case (passing a float to
a function expecting a long) is rare, and the much-more-common case
is we pass a long to these functions.

So first check if the type is long, and if it is then skip the expensive
PyFloat_Check check.

I didn't measure any speedup on the macrobenchmarks, but things like
`"".center(0)` are about 15% faster.


undingen (approved):
Good catch, didn't know that float does not have bitmask and has to do a slow check.

PR #167, opened 2022-01-24
JIT: a few small improvements I discovered while reviewing the method IC PR

undingen:
Only the 3 commits are new (I rebased it on your commit because I used the new `IS_IMMORTAL()` and I did not want to introduce a merge conflict:
- fix perf integration for some 'callq' instructions where we did not print the destination:
Before: 
image
After:
image

- jit: special case for moving an imm to mem without using a register
- jit: LOAD_GLOBAL don't emit an incref for immortal objects


kmod:
:+1:

PR #168, opened 2022-01-26
Cleanup the JIT code a bit for work on arm support

undingen:
No functional change.
Just some refactoring / removing disabled code.


kmod:
:+1:

PR #169, opened 2022-01-26
export SETUPTOOLS_USE_DISTUTILS=stdlib as a workaround for setuptools 60

undingen:
We are working on a permanent solution but in the meantime I thought we may want to workaround
it this way.
What do you think?


kmod:
Haha I kind of hate that we have to do this, but I think it's our best option for now.

PR #170, opened 2022-01-26
Reduce verbosity of `make install`

kmod:
This PR reduces the log output of `make install` from 10k lines to 350 lines.
I find the output annoying because it takes up my entire scrollback buffer, and it means that our github actions logs are so long that they cannot be displayed on github.com.

As a bonus I parallelized a bit of the installation process because on principle I'm annoyed to wait for a serial process on a 16-core machine haha


undingen (approved):
:+1:

PR #175, opened 2022-02-10
ARM64 preperations

undingen:
This PR refactors the JIT to make it easy to add support for ARM64.
Everything except the IC code is converted to use the newly introduced helper functions (used functions where possible and a few macros for branches where we have to). The generated code is slightly different for `STORE_FAST` and `DELETE_FAST` bytecodes and I fixed a bug in `END_FINALLY`.

I still have to cleanup the real arm implementations so this does not contain any of it yet but I hope to have it ready by next week (without the ICs). And plan on fixing the IC stuff afterwards.

It's best to look at every commit individually.

### DIFF
--- a/Python/aot_ceval_jit.c
+++ b/Python/aot_ceval_jit.c
@@ -214,7 +214,7 @@ struct PerfMapEntry {
     long func_size;
 } *perf_map_funcs;
 
-static int jit_use_aot = 1;
+static int jit_use_aot = 1, jit_use_ics = 1;
 
 static PyObject* cmp_outcomePyCmp_BAD(PyObject *v, PyObject *w) {
   return cmp_outcome(NULL, PyCmp_BAD, v, w);
@@ -1461,7 +1461,7 @@ static int emit_inline_cache(Jit* Dst, int opcode, int oparg, _PyOpcache* co_opc
     |  jne false_branch
     |.endmacro
 
-    if (co_opcache == NULL)
+    if (co_opcache == NULL || !jit_use_ics)
         return 1;
 
     // do we have a valid cache entry?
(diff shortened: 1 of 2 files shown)


Python/aot_ceval_jit.c
```diff
@@ -3215,15 +3215,22 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
...
 #endif
 
     size_t size;
-    dasm_link(Dst, &size);
+    int dasm_link_err = dasm_link(Dst, &size);
+    if (dasm_link_err) {
+#if JIT_DEBUG
+        fprintf(stderr, "dynasm_link() returned error %x", dasm_link_err);
+        JIT_ASSERT(0, "");
+#endif
```
kmod, line 3361: Just an idea -- if you extend JIT_ASSERT like we did with the old [RELEASE_ASSERT](https://github.com/pyston/pyston_v1/blob/v1.0/src/core/common.h#L53) then these four lines could just be `JIT_ASSERT(dasm_link_err == 0, "dynasm_link() returned error %x", dasm_link_err)`

Python/aot_ceval_jit.c
```diff
@@ -921,20 +1002,19 @@ static void emit_decref(Jit* Dst, int r_idx, int preserve_res) {
     emit_dec_qword_ptr(Dst, &_Py_RefTotal, 1 /* can_use_tmp_reg  */);
 #endif
     | add qword [Rq(r_idx)], -1
-
     // normally we emit the dealloc call into the cold section but if we are already inside it
     // we have to instead emit it inline
     int use_inline_decref = Dst->current_section == SECTION_COLD;
     if (use_inline_decref) {
-        | jnz >8
+        | branch_ne >8 // this is same as 'branch if not zero'
```
kmod: I always have to check whether ne is nz or z, what do you think about having a `branch_nz` macro as well?

kmod (approved):
Looks great!

undingen:
good idea!

PR #176, opened 2022-02-14
support building pyston on ARM64 without JIT

undingen:
initial JIT support will come in the next week or so.

### DIFF
--- a/Objects/stringlib/eq.h
+++ b/Objects/stringlib/eq.h
@@ -3,7 +3,7 @@
 /* Return 1 if two unicode objects are equal, 0 if not.
  * unicode_eq() is called when the hash of two unicode objects is equal.
  */
-Py_LOCAL_INLINE(int)
+__attribute__((always_inline)) Py_LOCAL_INLINE(int)
 unicode_eq(PyObject *aa, PyObject *bb)
 {
     PyUnicodeObject *a = (PyUnicodeObject *)aa;
--- a/pyston/nitrous/CMakeLists.txt
+++ b/pyston/nitrous/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ferror-limit=5 -fcolor-diagnostics")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ferror-limit=5 -fcolor-diagnostics")
 
-set(LLVM_LIB_DEPS LLVMCore LLVMSupport  LLVMBitReader LLVMAsmParser  LLVMTransformUtils LLVMScalarOpts  LLVMOrcJIT LLVMX86CodeGen LLVMX86AsmParser)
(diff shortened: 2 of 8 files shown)


Objects/stringlib/eq.h
```diff
@@ -3,7 +3,7 @@
 /* Return 1 if two unicode objects are equal, 0 if not.
  * unicode_eq() is called when the hash of two unicode objects is equal.
  */
-Py_LOCAL_INLINE(int)
+__attribute__((always_inline)) Py_LOCAL_INLINE(int)
```
kmod, line 6: Just curious, why do we need this?

pyston/nitrous/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ferror-limit=5 -fcolor-diagnostics")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ferror-limit=5 -fcolor-diagnostics")
 
-set(LLVM_LIB_DEPS LLVMCore LLVMSupport  LLVMBitReader LLVMAsmParser  LLVMTransformUtils LLVMScalarOpts  LLVMOrcJIT LLVMX86CodeGen LLVMX86AsmParser)
+set(LLVM_LIB_DEPS LLVMCore LLVMSupport  LLVMBitReader LLVMAsmParser  LLVMTransformUtils LLVMScalarOpts  LLVMOrcJIT LLVMAArch64CodeGen LLVMAArch64AsmParser)
```
kmod: Oh hmm, how is this passing the x86 github CI? Do we not use codegen anymore?

kmod (approved):
lgtm except I'm confused by how it works to always use the aarch64 libraries even on x86

pyston/nitrous/CMakeLists.txt
undingen: I took this commit from your arm branch. I does not work on x86 but I did add the arm/x86 check in my [3c623f5](https://github.com/pyston/pyston/pull/176/commits/3c623f5b27fa14ae1a5a30ce288a180096d40edf) commit.
(gettiing `undefined symbol: LLVMInitializeX86TargetInfo` without my fix)

undingen: image

kmod:
ahhh makes sense

Objects/stringlib/eq.h
undingen: I just rechecked this and I'm getting this error without it:
```
/usr/bin/ld: aot_optimized.o: in function `lookdict_unicode_nodummy':
/home/undingen/dev/pyston/build/bc_build/../../Objects/dictobject.c:892: undefined reference to `unicode_eq'
```
Seems like on ARM the compiler makes a bit different inlining decision.

undingen: And the problem is that this function is defined in a header which get included from `dictobject.c` and `unicodeobject.c` So just removing the static will make it complain about having it defined twice. (`Py_LOCAL_INLINE` is `static inline`)

kmod: Ah I see, yeah I think removing `static` has been our more typical approach so I wasn't sure why that wasn't used here, but now I get it

PR #177, opened 2022-02-16
Update to LLVM 13.0.1

kmod:
I was running into an optimization problem with one of our traces, and tracked it down to something that was fixed in newer versions of LLVM. So I decided to update our LLVM from 10.0 to 13.0.1 (latest release). The main issues:
- ORC v1 was removed, so I had to migrate to ORC v2
- We had some bugs that were not triggered by the IR that LLVM 10.0 generates, but are triggered by LLVM 13

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:20.04
 
 ENV DEBIAN_FRONTEND=noninteractive
 
--- a/pyston/pystol/pystol.cpp
+++ b/pyston/pystol/pystol.cpp
@@ -262,8 +262,7 @@ bool PystolFactDeriver::deriveFacts(Value* v, FactSet& facts, LLVMEvaluator& eva
             } else if (name == "PyTuple_New" || name == "PyTuple_New_Nonzeroed") {
                 RELEASE_ASSERT(isPyObjectPtr(call->getType()), "");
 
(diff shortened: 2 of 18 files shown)


kmod:
The PR didn't trigger CI so I tried marking it "ready for review", but the issue was that there was a merge conflict. I don't see how to set it back as a draft -- this isn't actually ready for review.

kmod:
Ok this is now ready for review!

.github/workflows/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:20.04
```
undingen, line 1: good catch!

pyston/pystol/pystol.cpp
```diff
@@ -262,8 +262,7 @@ bool PystolFactDeriver::deriveFacts(Value* v, FactSet& facts, LLVMEvaluator& eva
             } else if (name == "PyTuple_New" || name == "PyTuple_New_Nonzeroed") {
                 RELEASE_ASSERT(isPyObjectPtr(call->getType()), "");
 
-                Type* t = call->getType();
-                t = call->getParent()->getParent()->getParent()->getTypeByName("struct._typeobject")->getPointerTo();
+                Type* t = StructType::getTypeByName(getContext(), "struct._typeobject");
```
undingen, line 265: :smile:

undingen (approved):
yeah very nice :sparkler:

PR #178, opened 2022-02-17
jit: add ARM64 support (without ICs) 

undingen:
Passes all tests and testsuites but does not implement Inline caches yet which will follow next week.
Also codegen is not optimized yet for ARM.

Most of the work has already been factored out in the previous pull requests now we just
have to actually implement the ARM codegen.
Luckily it's not that much code and most of it is related to the fixed 32bit instructions size
which is much more strict about the size of immediates.

I guess the most controversial thing will be my decision to implement everything in the same file and use a custom
preprocessor. I decided on this route because we can reuse most of the functions and often the arch specific code is only a few lines and most stuff is shared.
So to me it makes not much sense to split the archs into different files / I actually even prefer to see what the other arch is doing.
Unfortunately this would be really verbose because it always requires us to wrap everything into a C ifdef and a DynASM ifdef.
More infos regarding this are found in the preprocessor source.

__There a some notable gotchas:__
- inline caches and `Py_REF_DEBUG` are not yet implemented
- on ARM64 the return value of a function is the same reg as the first argument. This causes a bunch of issues for us.
I worked around it by assigning reg 'res' to new register and making sure that we generate a extra mov after
every call/return instruction which moves the value from 'real_res' to 'res'.
This means we generate a few unnecessary `mov reg, reg`s but this should be super cheap and we can fix it later if we want.
- we can't compile functions with more than 16383 bytecodes (this limit will double when we upgrade to cpython 3.11)
This makes the `f_lasti` update code easier and I don't think this is a problem we will just abort compilation for this specific function.
- `mmap()` does not support `MAP_32BIT`. I worked around it by trying to allocate fixed addresses which seems to work fine even to it's a bit of hack.

**Performance:**
I run a quick pyperformance comparison between Ubuntu 20.04 CPython and Pyston.
For the Intel run I disabled ICs and BOLT to make the comparison fairer.
All are using the optimized build.
```
Geometric mean speedup of Pyston with disabled Inline Caches vs CPython 3.8.10:
M1 Pro:          1.28x faster
Graviton2:       1.33x faster
RPi4:            1.36x faster
Intel i9-12900K: 1.41x faster - x86_64 just for comparison
```
The M1 was the only machine where a few tests performed significantly slower than CPython not sure if this is just a fluke -
will investigate it later.

### DIFF
--- a/pyston/tools/dynasm_preprocess.py
+++ b/pyston/tools/dynasm_preprocess.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+#
+# Preprocesses the DynASM file to remove architecture specific code depending on local architecture.
+# Non matching architecture code is replaced with empty lines so that line numbers match.
+# Usage:
+#   python3 dynasm_preprocess.py <input_file> <output_file>
+#
+# Currently supported architectures:
+#   ARM -> aarch64
+#   X86 -> x86_64
+#
+# Supported directives:
+#  Directives must be first nonwhite space characters on the line.
+#  @<ARCH><CODE> # removes <CODE> depending on arch
+#  @<ARCH>_START # removes all lines to the next @<ARCH>_END depending on arch
+#  @<ARCH>_END
+#
+# Example single line directive:
+#  @ARM| .define name, reg_arm64
+#  @ARM|| #define name_idx reg_arm64_idx
+#  @X86| .define name, reg_amd64
+#  @X86|| #define name_idx reg_amd64_idx
+#
+# Example multi line directive:
+#   @X86_START
+#       if (val == 0) {
+#           | test Rd(r_idx), Rd(r_idx)
+#       } else if (IS_32BIT_VAL(val)) {
+#           | cmp Rd(r_idx), (unsigned int)val
+#       } else {
+#           JIT_ASSERT(0, "should not reach this");
+#       }
+#   @X86_END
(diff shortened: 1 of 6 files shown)


undingen:
I just checked if it's still working after rebasing it on the LLVM 13 commits and I can confirm that it's working

undingen:
Unfortunately there is a problem on ARM with generating the PGO info during the opt build. I'm currently investigating

pyston/tools/dynasm_preprocess.py
```diff
@@ -0,0 +1,120 @@
...
+            if line_stripped.startswith("@"):
+                # we go over all supported archs because it makes it easier to catch errors
+                for arch in ARCHS.values():
+                    directive = "@" + arch
+                    if line_stripped.startswith(directive):
+                        directive_start = directive + "_START"
+                        directive_end = directive + "_END"
+
+                        if line_stripped.startswith(directive_start):
+                            inside_multiline_directive = directive_start
```
kmod, line 90: Maybe add an `assert inside_multiline_directive is None` here?

kmod (approved):
This looks amazing!

pyston/tools/dynasm_preprocess.py
undingen: good point, I added a error message

PR #179, opened 2022-02-22
arm64: run PGO tasks single processed

undingen:
I'm seeing a lot of this errors on the `aot_pic` profiles of the `optshared` build on ARM (`opt` seems to work):
```
LLVM Profile Error: Invalid profile data to merge
LLVM Profile Error: Profile Merging of file pyston/build/aot_pic/aot_prof/default_16542605575556428325_0.profraw failed: File exists
```
When running manually `llvm-profdata` on it I'm seeing a lot of truncation and invalid header errors.

I tried adding the process ID into the filename of the profile file and merge it only afterwards but this does not fix the problem.
Also looked at the bugtracker but could only find a similar sounding issue for windows which got already fixed a long time ago.
In general multiprocessed PGO runs should work - I also creating a testcase and it worked fine.
Not sure whats going on here - it works fine on x86 and I think it's new since llvm 13.

I think we should just run it single processed on ARM for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,12 @@ endif
 CPYTHON_EXTRA_CFLAGS:=-fstack-protector -specs=$(CURDIR)/pyston/tools/no-pie-compile.specs -D_FORTIFY_SOURCE=2
 CPYTHON_EXTRA_LDFLAGS:=-specs=$(CURDIR)/pyston/tools/no-pie-link.specs -Wl,-z,relro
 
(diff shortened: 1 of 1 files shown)


kmod:
:+1:

undingen:
Unfortunately I noticed now they same message in the https://github.com/pyston/pyston/pull/183 CI x86 release build. 
Will take a look again

PR #181, opened 2022-02-23
Some minor optimizations

kmod:
A bunch of misc minor things that I'd noticed over time. The net effect is small, looks like about 0.1%

I'm somewhat optimistic that the immortalizing stuff will have more benefit as we find more ways to use it.

### DIFF
--- a/pyston/aot/aot_gen.py
+++ b/pyston/aot/aot_gen.py
@@ -109,6 +109,7 @@ class Signature(object):
         self.name = "".join([cls.name for cls in argument_classes]) + str(self.nargs)
         self.always_trace = list(always_trace)
         self.guard_fail_fn_name = guard_fail_fn_name
+        self.do_not_trace = []
 
     def getGuard(self, argument_names):
         names = list(argument_names)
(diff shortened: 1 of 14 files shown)


pyston/aot/aot_gen.py
```diff
@@ -450,7 +450,7 @@ def _args_names(self):
 
     def _args(self):
         args = [("PyThreadState *", "tstate"),
-                ("PyObject **", "stack"),
+                ("PyObject ** restrict", "stack"),
```
undingen, line 492: nice that this small change fixes the issue with the new llvm :)

pyston/aot/aot_gen.py
```diff
@@ -430,6 +452,15 @@ def write_profile_func(self, traced, header_f, profile_f):
...
         print("PyObject* f = *(stack - oparg - 1);", file=profile_f)
 
+        # We want to make sure that we try more-specialized traces before trying
+        # the less-specialized ones. In the past we carefully ordered the creation
+        # of the traces so that they would be tested in the same order that they
+        # would were created, but now that there are some constraints around the
+        # order in which we can create them (have to create the less-specialized first),
+        # use an explicit measure of specialization to order them.
+        traced = list(traced)
+        traced.sort(key=lambda p: -p[0].getSpecializationLevel())
```
undingen, line 470: very good catch! I would not have noticed :(

undingen (approved):
LGTM

PR #182, opened 2022-02-24
opt build: set -vp-counters-per-site

undingen:
I'm seeing a lot of:

>LLVM Profile Warning: Unable to track new values: Running out of static counters.  Consider using option -mllvm -vp-counters-per-site=<n> to allocate more value profile counters at compile time.

Also seeing it in the CI logs.
Increasing it to 2 seems to fix the issue.

clang itself is also increasing it so should be fine: https://reviews.llvm.org/D92669


kmod:
:+1:

PR #184, opened 2022-03-10
Update bolt repo to llvm now that it's merged

kmod:
And update to the latest commit

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "pyston/bolt/bolt"]
 	path = pyston/bolt
-	url = https://github.com/pyston/BOLT
+	url = https://github.com/llvm/llvm-project
 [submodule "pyston/LuaJIT"]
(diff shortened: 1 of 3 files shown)


undingen:
👍, when you update please also update our conda bolt recipe (and upload a new bolt conda package) because else the CI will keep on using the older bolt version.

kmod:
Ok looks like CI passed, this is ready for review now

undingen (commented):
already wanted to press merge but now I noticed the following:
I think the `build_num` in `pyston/conda/pyston/meta.yaml` needs to be incremented. Because the version number of the pyston package didn't get touched. (the bold package should be fine because it's version number changed).
I don't think we plan to release a new pyston package currently so it's theoretically only a problem...
Else looks good :)

kmod:
Ah ok that's a good idea, I went ahead and did that

PR #185, opened 2022-03-13
make sure our global builtin objects have a GC header

undingen:
I caught this when compiling pyston via clang (caused many segfaults).
Problem is that the `PyCFunctions` have `Py_TPFLAGS_HAVE_GC` set so must have been allocated via `PyObject_GC_New()`.
I switched to allocate them as global variable because this allows our traces to be further optimized but
I didn't allocate space for the GC Header in front of the object.

This should fix it - performance and generated code seems to be fine.
I checked that `_PyObject_GC_New` -> `_PyObject_GC_Alloc` is setting the GC header to 0 so this should now be the same as we do here.


kmod:
:+1:

Can't believe this didn't show up before!

PR #186, opened 2022-03-14
provide conda packages for arm64

undingen:
- removed the compiler-rt package because conda-forge already provides it in the correct version
- don't build bold because it seemed to not work on arm64 currently
- use luajit via apt-get because conda-forge provides only x86 packages and lua misses a c extension
- removed 'target_platflorm' inside variants.yaml with it would overwrite linux-aarch64 to linux-64
  I think this should now work for both platforms and we don't need two separate files.


undingen:
Ok so I updated the patch but I noticed now that everything does seem to work but the conda build executable is a bit slower than the non conda one. It seems e.g. for nbody.py I see a lot of `__inf` calls which are inlined in the non conda build (7.1s vs 7.8sec). I investigate this a bit more tomorrow and suggest to not build 3rd party packages yet because I may have to modify the CFLAGS.

undingen:
Ok so I sneaked in a commit which brings the conda build pyston performance close to the one build via ubuntu 20.04 toolchain.
If the CI turns green it should be ready for review.

kmod:
awesome!

PR #187, opened 2022-03-21
provide ubuntu packages for arm64

undingen:
- some special cases because we don't use bolt on arm
- run into a compilation error on 18.04 when compiling asan in compiler-rt - disabled it

I only did limited testing: 
- tried installing the 20.04 package and installed and run numpy
- run some simple benchmarks using the portable release


kmod:
:+1:

PR #189, opened 2022-03-27
jit: use rip relative call instruction 'bl' on Arm64

undingen:
This replaces the very often used call sequence:
```ASM
mov tmp2, func_addr&0xFFFF
mov tmp2, func_addr>>16, lsl #16
blr tmp2
```
with:
```ASM
bl func_addr
```
folding 3 instructions into 1.

Unfortunately bl can only address offsets +-128MB. This means we have to be very careful when allocating memory for the
the JIT code. Because we don't know yet during code generation where the code will actually be placed (but
also even if we would it will be patched with a different addr).
I decided to just use a heuristic that we want to be +-64MB next to one of the AOT funcs which should make
sure we can patch the AOT func call to a different func.

I had to leave in a path for functions which can't be called this way for calls generated by the `CALL_METHOD` ICs
to external functions in C extensions which may be further away but can't get patched via `SET_JIT_AOT_FUNC`.

Results in an impressive 4% speedup on M1 pro pyperformance. On graviton2 performance is only 1% improved.
It seems to decrease number of branch misses a lot so there may be some limitations in the CPU with the number
of absolute calls it can track or so which this fixes.

We could also use this approach later on x86 where it would help the shared build where we often have to fallback to the absolute call instruction. (this is not an issue form the normal build where we can use the +-2GB relative call instruction to reach everything)


kmod:
:+1:

PR #190, opened 2022-03-28
dynasm: fix |.align issue on arm64

undingen:
Update to latest git revision which contains a fix for `|.align` on arm64
which used to generate invalid code (used wrong value for a NOP instruction on arm64).

We are currently not executing the instructions in `|.align` (we only use it to align the start of the function)
so this won't change things but I did run into it when I experimented with adding align directives to jump targets.

Also seems like nothing else changed inside DynASM (remaining changes are LuaJit related which we don't use).


kmod:
:+1:

PR #191, opened 2022-03-28
misc jit speedups

undingen:
- the arm64 changes not only increase speed but also reduce code size
- most "controversial" is likely the "remove signal and tracing check for STORE_FAST" change. It helps especially unpacking code. All tests pass but can't guarantee it's not breaking some 3rd party tracing code.. I'm fine with removing it if you have doubts.

Speedup is about 4% on Graviton2 and RPi4 (and less on M1) on pyperformance.
```
Geometric mean pyperformance speedup compared to ubuntu 20.04 cpython
                    #183                this patch
M1 Pro:          1.38x faster      ->  1.45x faster
Graviton2:       1.42x faster      ->  1.48x faster
RPi4:            1.46x faster      ->  1.55x faster
Intel i9-12900K: 1.50x faster                             - x86_64 just for comparison
```

### DIFF
--- a/Python/aot_ceval.c
+++ b/Python/aot_ceval.c
@@ -4482,13 +4482,6 @@ _PyEval_EvalFrame_AOT_JIT(PyFrameObject *f, PyThreadState * const tstate, PyObje
 {
     PyObject* retval = NULL;
 
-    // our generated code dispatches based on f_lasti so we need to adjust it
-    if (f->f_lasti < 0)
-        f->f_lasti = 0;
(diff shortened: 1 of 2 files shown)


kmod:
Looks good! I wonder if there are other places that we can avoid doing the signal/tracing checks? I don't remember exactly what's specified to happen but maybe at some point we could systematically go through and figure out which ones we need to do.

undingen:
yes that's a good idea I'm quite sure we will find additional places

PR #192, opened 2022-03-29
docker: create multi arch images using buildx

undingen:
This seems to be the recommended way to create this images. And allows via qemu user mode emulation to generate all images on the same machine.

I tested a dummy release on my private repo (with test packages so please don't use this for anything) and
could push it successfully to `undingen/pyston`.
Afterwards I docker run it fine on my x86 PC and MacOS M1 ARM.

I decided to use the debian arch names (amd64,arm64) as the suffix for the debian packages.
And the conda default ones (x86_64,aarch64) for the installers like e.g. similar to the miniconda installer.

### DIFF
--- a/pyston/docker/build_docker.sh
+++ b/pyston/docker/build_docker.sh
@@ -1,18 +1,22 @@
 #!/bin/sh
+#
+# Requires:
+# - docker buildx
+# - installed qemu emulation for the non native arch
+#   e.g.: docker run --privileged --rm tonistiigi/binfmt --install arm64
 
 set -eux
 
 BUILD_NAME=2.3.2
 DIR=$(dirname $0)
+PLATFORMS=linux/amd64,linux/arm64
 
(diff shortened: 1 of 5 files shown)


pyston/docker/build_docker.sh
```diff
@@ -1,18 +1,22 @@
 #!/bin/sh
+#
+# Requires:
+# - docker buildx
+# - installed qemu emulation for the non native arch
+#   e.g.: docker run --privileged --rm tonistiigi/binfmt --install arm64
```
kmod, line 6: Can you add these requirements to the readme (section "Extra dependencies for producing Pyston debian packages and portable directory release")?

kmod (approved):
looks great!

pyston/docker/build_docker.sh
undingen: good point - documented it

PR #195, opened 2022-04-06
aot: use musttail

undingen:
now that we upgraded to clang 13 we can remove this workaround which should also be a tiny speedup.
Problem was that if we call (instead of jump) from a AOT func to another AOT func
and this one calls SET_JIT_AOT_FUNC() it would look at the callers address which would not point
to the JIT compiled func but to the other AOT func.
By using tail calls (=jumps) this issue is gone.

Clang is now per default already generating tail calls in this cases so I can't reproduce the crash even without
specifying it but I verified that the generated code looks good.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,12 +138,12 @@ build/systemdbg_env/bin/python: | $(VIRTUALENV)
 build/pypy_env/bin/python: | $(VIRTUALENV)
 	$(VIRTUALENV) -p pypy3 build/pypy_env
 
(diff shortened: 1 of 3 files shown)


undingen:
it seems like bolt is getting rid of the tailcalls? which causes problems. I may close this issue if skipping BOLT on AOT funcs hurts perf.

kmod:
Ah that's too bad about bolt. I actually don't think your original solution is that bad...

PR #198, opened 2022-04-11
jit: use PyUnicode_Append when possible

undingen:
This allows us to modify the string inplace if the string is only stored in the fast array
and the operation is followed by a STORE_FAST which will overwrite it.

`PyUnicode_Append()` will handle cases where the reference count is 1 or more (the interpreter also always calls this
function for unicode objects without doing any reference count check first).

The interpreter was already doing it but the JIT didn't implement it.
Code like this gets about 20% faster:
```python
    a = ''
    for i in range(200):
        a = a + 'a'
```


kmod:
lgtm!

It looks like CPython can also handle cases like `x = make_string() + make_string2()`, ie another case where the lhs might have a refcount of only 1. Do you think it's worth handling this too?

undingen:
very good point I overlooked that. Will investigate and add a new PR.

PR #199, opened 2022-04-12
add a STORE_ATTR cache for attributes in __slots__

undingen:
this cases should be common and is easy to create a IC for.
I noticed it while investigating performance of scons.
Loop which just sets an attribute is 2x faster with it
Our large benchmarks don't really change perf
(except 'float' from pyperf gets faster because it has lots of slots access).


kmod:
this is great! maybe we can encourage people now to use slots more since it seems like one of the fastest and most reliable caching mechanisms.

undingen:
Yes slots should be super fast also CPython 3.11 has same fast caches for it:
https://github.com/python/cpython/blob/54f67ad54366f1b9161edca283e19670e065e1b8/Python/ceval.c#L3636

PR #200, opened 2022-04-13
perf: display variable names, attributes etc

undingen:
Makes it much easier to compare python source with generated machine code.
Displays now also on the first line the filepath to make it easy to open it in an editor.

Theoretical we can also mix in the python code later with the disassembly.


kmod:
:+1:

PR #201, opened 2022-04-18
 jit: special case loading methods from constants

undingen:
This mainly helps `"".join()` and `"".format()`.
Don't expect any noticeable perf difference this was just something I noticed while looking at other stuff.


kmod:
:+1:

PR #202, opened 2022-04-18
obmalloc: prefer 32bit addresses

undingen:
per default all heap allocated python objects have >32bit address on my machine.
With this change they are <32bit on x86 which allows the JIT to generate slightly smaller code.

Tradeoff is that I guess ASLR will be less random otherwise I don't see any disadvantages.
I'm slightly in favor of merging it but don't feel strongly about it. I did not measure a performance on our larger tests.


kmod:
:+1:

PR #203, opened 2022-04-19
jit: inline list or tuple indexing with fixed index

undingen:
even though we already have AOT traces for GetItem this is still a bit faster for this special cases.
E.g. speedups nbody by about 10-15%.

### DIFF
--- a/Include/internal/pycore_code.h
+++ b/Include/internal/pycore_code.h
@@ -124,9 +124,14 @@ typedef struct {
     char cache_type;
 } _PyOpcache_StoreAttr;
 
+typedef struct {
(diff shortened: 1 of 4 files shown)


undingen:
I'm not happy about our `PyObject_GetItemLongList3` trace. 
While the only difference to the list one is that there is a extra:
```C
if (PyUnicode_READY(self) == -1)
        return NULL;
```
check in the code it seems like llvm can't figure out anymore that `PyIndex_Check(item)` is true. While it can for the list.

kmod:
Looks good! I'm still surprised that this is that much better than the aot traces, seems like maybe those could be better.

PR #205, opened 2022-04-26
re: speedup matching by storing some state in local variables

undingen:
This is a backport of the upstream 3.11 improvement:
https://github.com/python/cpython/pull/91495

I only backported the `ctx->pattern` -> `pattern` and `ctx->ptr` -> `ptr`
part because using computed goto actually decreased perf slightly on the opt build.

Speedup on the pyperf regex benchmarks is more like 3-5% on the opt build.


kmod:
cool!

PR #206, opened 2022-04-26
make_release: speedup release build, run git cleanup

undingen:
- don't use docker to copy files instead mount it and make a copy which is a huge speedup
- "Sending build context to Docker daemon  17.36GB" is reduced to a few KB by moving the Dockerfiles into new directory.
- run `git clean -fdx` to make sure we don't accidentally include some unintended stuff

I made this changes while debugging the ubuntu 22.04 packaging issue because I got annoyed about the huge build overhead.


pyston/tools/make_release.sh
```diff
@@ -57,9 +58,16 @@ function make_release {
...
 
     echo "Creating $DIST release"
-    docker build -f pyston/Dockerfile.$DIST -t pyston-build:$DIST . --no-cache
-    docker run -iv${PWD}/release/$VERSION:/host-volume --rm --cap-add SYS_ADMIN pyston-build:$DIST sh -s <<EOF
+    docker build -f pyston/make_release_files/Dockerfile.$DIST -t pyston-build:$DIST pyston/make_release_files/ --no-cache
+    docker run -iv${PWD}/release/$VERSION:/host-volume -v${SRC_DIR}:/src/src_dir_host:ro --rm --cap-add SYS_ADMIN pyston-build:$DIST sh -s <<EOF
 set -ex
+
+# make a copy of the source because it's mounted read only and we want to modify it
+cp -r /src/src_dir_host /src/build
```
kmod, line 66: minor improvement but it looks like rsync can take an `--exclude` argument to not copy the build dircetory at all
https://stackoverflow.com/questions/4585929/how-to-use-cp-command-to-exclude-a-specific-directory

PR #207, opened 2022-05-01
jit: STORE_ATTR inline cache for split dicts

undingen:
Unfortunately it's not really possible to inline all of it
because of the complexity of setting a dict entry so we still branch
out to a helper function at the end.

But it's a bit faster then not having the cache inline.
make measure perf stays the same microbenchmarks are slightly improved (e.g richards).
Cache hit rates are really really good for most benchmarks I looked at.
```
e.g. both of this couldn't use an IC for STORE_ATTR before:
raytrace: inlined 10 (of total 10) STORE_ATTR caches: 16015419 hits 0 misses (=0%)
richards: inlined 41 (of total 54) STORE_ATTR caches: 27436324 hits 629121 misses (=2%)
```

I hope that in the future with the cpython 3.11 object layout for split dicts this caches
should be faster and smaller because the pointer to the values array is directly stored
in front of the object.

I confirmed that it also works on Arm.

### DIFF
--- a/Include/internal/pycore_code.h
+++ b/Include/internal/pycore_code.h
@@ -105,10 +105,11 @@ enum _PyOpcache_StoreAttr_Types {
 
     // caching an index inside instance splitdict, guarded by the splitdict keys version (dict->ma_keys->dk_version_tag)
     SA_CACHE_IDX_SPLIT_DICT = 0,
+    SA_CACHE_IDX_SPLIT_DICT_INIT = 1, // same as the first but means we hit the dict not initialized path
 
     // caching the offset to attribute slot inside a python object.
     // used for __slots__
-    SA_CACHE_SLOT_CACHE = 1,
+    SA_CACHE_SLOT_CACHE = 2,
 };
 
 typedef struct {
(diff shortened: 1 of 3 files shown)


kmod:
lgtm, but I had a thought -- for any given STORE_ATTR ic, it's essentially either going to always be the first attribute or never be the first attribute. Do you think it's worth including that in the cache and writing it out slightly differently ICs for the two cases?

undingen:
I'm not sure if I understood you correctly but I experimented now with assuming that the split dict index is 0 in the init path.
But hit several cases in the testsuite where this assumption is not true (not sure about how important it is in the benchmarks).
E.g code like this which with this PR always hits the cache:
```python
 def f(x):
    x.z = 1 # will create an IC with split dict index 1
class C:
    pass
def g():
    c = C()
    c.a = 2 # index 0
    c.z = 3 # index 1
    for i in range(100000):
        f(c) # existing dict
    f(C()) # has to init dict from the shared keys
g()
```

undingen:
I have put up a new version which only emits one path. Either for initializing a dict or modifying a existing one. Like you said the statistics show that its very uncommon that a ICs uses both paths.
Also it's now the other way around: I rebased this PR on top of #208 because it's already reviewed and will go in first.

kmod:
:+1:

PR #208, opened 2022-05-01
ICs: cache tp_dictoffset

undingen:
If we guard on `tp->tp_version_tag` we can also assume that the `tp_dictoffset` has not changed.
This removes a load, cmp, conditional branch for most IC entries
it also allows us to identify types with negative `tp_dictoffset` at compile time which we currently don't handle in the JIT.

I choose to use a short to store the offset because it will not increase the size of our caches
and it should handle nearly all cases.

I'm measuring a small improvement on our benchmarks and pyperformance.

This PR is currently on top of `STORE_ATTR` because it also modifies this cache so it's best to review #207 before.

### DIFF
--- a/Include/internal/pycore_code.h
+++ b/Include/internal/pycore_code.h
@@ -94,6 +94,7 @@ typedef struct {
             int64_t offset; /* offset in bytes from the start of the PyObject to the slot */
         } slot_cache;
     } u;
+    short type_tp_dictoffset;  /* tp_dictoffset of type */
     char cache_type;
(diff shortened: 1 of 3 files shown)


undingen:
Found a bug with the `version_zero` case in the previous commit of this PR (which made us miss a few cases which we previously succeeded at caching).
Should be ready for review when it turns green.

kmod:
lgtm!

undingen:
I have put up a new version which only removes one path. Either for initializing a dict or modifying a existing one. Like you said the statistics show that its very uncommon that a ICs uses both paths.

kmod:
:+1:

PR #209, opened 2022-05-02
make_release: use rsync with --exclude build

undingen:
this way we don't have to copy this large directory just to erase it immediately afterwards.
Fixes review comment in #206


kmod:
lgtm

PR #211, opened 2022-05-04
add a brief history of the project to the readme

kmod:
since we often get questions about it or about where the LLVM tier is


undingen (approved):
:+1:

PR #212, opened 2022-05-05
ICs: add polymorphic LOAD_ATTR/LOAD_METHOD caches

undingen:
Add a new cache type which stores a dynamically allocated array to further opcache entries.
This are mostly useful for the JIT which can emit several entries with low overhead.

I had to come up with some heuristics to not generate too many polymorphic sites
and what seemed to work was to check if the type is the same (we us a hash because
of storage requirements in the cache entry). This seems to catch a lot of cases
where the shape is quickly changing. In addition the JIT will only emit entries
which at least once got executed which also catches some bad entries.

This improves our benchmarks but also pyperformance (especialy 'richards': 1.39x faster)

Cache entries will never be freed so will leak but I this is similar
to how JIT memory is already leaking.


kmod:
This is way more elegant than I was expecting!

PR #213, opened 2022-05-05
 Take over _PyEval_EvalFrameDefault instead of using PEP523

kmod:
_PyEval_EvalFrameDefault is not safe to call once Pyston is
enabled, due at least to the opcaches being different.

This function isn't part of the public API, but it's still somewhat
exposed, and in general it feels like an unnecessary risk to have
a function that will cause bugs if called.


undingen (approved):
good idea

PR #214, opened 2022-05-05
Change this from a 64-bit compare to a 32-bit one

kmod:
I think this is right?


undingen (commented):
Good catch - this is quite embarrassing.
I think it should just be `cmp dword [Rq(r_mem)+ offset], (unsigned int)val` because
the pointers addr is still 64bit so `Rq(r_mem)` and not `Rd(r_mem)`

Issue #19, opened 2021-05-07
Fedora/CentOS/Redhat/Oracle Linux packaging


woodgrinder:
Adding an RPM package would greatly increase the user base for pyston.

I know it's early days, but just want to put it on your enhancement list.

I think that once you get your user base expanded, you will have a real winner!

diego-garro:
don't forget `openSUSE` please :)

brymer-meneses:
Add Debian as well

undingen:
We would be interested in contributions which add support for further distros.
The packages are currently build inside a docker container see: https://github.com/pyston/pyston/blob/pyston_master/pyston/Dockerfile.20.04 and https://github.com/pyston/pyston/blob/pyston_master/pyston/tools/make_release.sh
If somebody knows how to add support for fedora and has some time available... :smile:

kmod:
@woodgrinder does the portable package (.tar.gz) work for you? Our hope was that it would be relatively distro-agnostic.

woodgrinder:
I have not had time to investigate.  Our environment is not connected to the internet and has really strict security, so there’s quite a bit of overhead.

I will look at it, when I can.

Thanks,

John

On May 11, 2021, at 9:48 AM, Kevin Modzelewski ***@***.******@***.***>> wrote:



@woodgrinder<https://github.com/woodgrinder> does the portable package (.tar.gz) work for you? Our hope was that it would be relatively distro-agnostic.

PR #217, opened 2022-05-10
Initial version of pyston_lite

kmod:
This is only a few percent faster, but I wanted to put up an initial PR and will add some optimizations in later PRs.

I originally did the development by copying all the jit files into a separate directory, but this made for a tough merge. I think it makes sense to try to keep just a single version of our jit with appropriate ifdefs for pyston_lite, so that's how this PR does it.

This PR is just the basics to get pyston_lite working, which is mainly removing the features that we can't use, and finding replacements for the ones that we still need. From a perf standpoint the biggest hit is the removal of the split-dict caches, but I will add my new version of the cache that helps with that.

In a future PR I will also make this a bit more safe in case you run for a while before switching to pyston_lite (in particular, dealing with any opcaches that were already initialized).

I used a bare-bones Makefile for now; you can test it by going into `pyston/pyston_lite` and doing `make bench` or `make bench_baseline`. I'll migrate this to a setup.py in a later PR

### DIFF
--- a/Python/aot_ceval.c
+++ b/Python/aot_ceval.c
@@ -1474,8 +1474,11 @@ typedef JitRetVal (*JitFunc)(PyFrameObject* frame, PyThreadState * const tstate,
 
 #ifdef PYSTON_LITE
 JitFunc jit_func_lite(PyCodeObject* co, PyThreadState* tstate);
+#define jit_func jit_func_lite
 void jit_start_lite();
+#define jit_start jit_start_lite
 void jit_finish_lite();
+#define jit_finish jit_finish_lite
 #else
 JitFunc jit_func(PyCodeObject* co, PyThreadState* tstate);
 void jit_start();
@@ -1485,6 +1488,28 @@ static long opcache_min_runs = OPCACHE_MIN_RUNS;
 static long jit_min_runs = JIT_MIN_RUNS;
 
 #define JIT_FUNC_FAILED ((JitFunc)0x1)
+
+#ifdef PYSTON_LITE
+static inline void* getJitCode(PyCodeObject* code) {
+    // _PyCode_GetExtra writes jit_code in all non-internal-error cases
+    void* jit_code = JIT_FUNC_FAILED;
+    _PyCode_GetExtra((PyObject*)code, code_jitfunc_index, &jit_code);
+    return jit_code;
+}
+
+static inline void setJitCode(PyCodeObject* code, void* jit_code) {
+    _PyCode_SetExtra((PyObject*)code, code_jitfunc_index, jit_code);
+}
+
+#else
+static inline void* getJitCode(PyCodeObject* code) {
+    return code->co_jit_code;
+}
+
+static inline void setJitCode(PyCodeObject* code, void* jit_code) {
+    code->co_jit_code = jit_code;
+}
+#endif
 
 static PyObject* _Py_HOT_FUNCTION
 _PyEval_EvalFrame_AOT_JIT(PyFrameObject *f, PyThreadState * const tstate, PyObject** stack_pointer, JitFunc jit_code);
@@ -1885,7 +1910,6 @@ _PyEval_EvalFrame_AOT_Interpreter(PyFrameObject *f, int throwflag, PyThreadState
 #endif
 
 // increments the number of times this loop got ececuted and if the threshold is hit JIT func and do OSR.
-#ifdef PYSTON_LITE
 #define HANDLE_JUMP_BACKWARD_OSR() \
     do {  \
         ++co->co_opcache_flag;  \
@@ -1893,12 +1917,11 @@ _PyEval_EvalFrame_AOT_Interpreter(PyFrameObject *f, int throwflag, PyThreadState
         /* check if we should switch over to the JIT (OSR) */  \
         if (co->co_opcache_flag > jit_min_runs && can_use_jit \
             && !_Py_TracingPossible(ceval)) { /* don't OSR if tracing is enabled because we seem to skip a line */ \
-            void* code = JIT_FUNC_FAILED; \
-            _PyCode_GetExtra((PyObject*)co, code_jitfunc_index, &code); \
+            void* code = getJitCode(co); \
             if (code == NULL) { \
-                code = jit_func_lite(co, tstate);  \
+                code = jit_func(co, tstate);  \
                 if (code) {  \
-                    _PyCode_SetExtra((PyObject*)co, code_jitfunc_index, code); \
+                    setJitCode(co, code); \
                     /* JUMPTO() did not update f->f_lasti  \
                     (it still points to the JUMP_ABSOLUTE - not the destination of the jump)  \
                      update f->f_lasti manually like DISPATCH() would do because  \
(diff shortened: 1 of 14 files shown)


kmod:
I just noticed a crash when running the cpython testsuite, I'll take a look into it

kmod:
ok I fixed the crash and this is ready for review

pyston/pyston_lite/aot_ceval_jit_helper.h
```diff
@@ -0,0 +1 @@
+../../Python/aot_ceval_jit_helper.h
```
undingen: This file does not exist

Python/aot_ceval.c
```diff
@@ -1838,30 +1943,52 @@ _PyEval_EvalFrame_AOT_Interpreter(PyFrameObject *f, int throwflag, PyThreadState
...
 
-    if (can_use_jit && co->co_jit_code == 0 && co->co_opcache_flag >= jit_min_runs /* jit after that many calls or gen yields */) {
-        // JIT assumes opcache is always on
-        if (co->co_opcache_map == NULL) {
-            _PyCode_InitOpcache(co);
-        }
+    if (can_use_jit && co->co_opcache_flag >= jit_min_runs /* jit after that many calls or gen yields */) {
+#ifdef PYSTON_LITE
+        void* code = JIT_FUNC_FAILED;
+        _PyCode_GetExtra((PyObject*)co, code_jitfunc_index, &code);
```
undingen: I find all this ifdefs and code duplication just to set and retrieve the the pointer to the machine code a bit ugly. What do you think about adding a helper function/macro which handles it?!?

Python/aot_ceval_jit.c
```diff
@@ -847,7 +900,11 @@ static void switch_section(Jit* Dst, Section new_section) {
 // compares r_type_idx->tp_version_tag with type_ver
 // branches to false_branch on inequality else continues
 |.macro type_version_check, r_type_idx, type_ver, false_branch
-|| emit_cmp64_mem_imm(Dst, r_type_idx, offsetof(PyTypeObject, tp_version_tag), (uint64_t)type_ver);
+#ifdef PYSTON_LITE
+|| emit_cmp32_mem_imm(Dst, r_type_idx, offsetof(PyTypeObject, tp_version_tag), (unsigned int)type_ver);
```
undingen, line 923: I had totally forgot about the difference in size of `tp_version_tag`, good that you noticed!
But the #ifdef/else/endif needs "||" infront. Right now dynasm generates both paths because it just seems to ignore them:
```C
    } else {
        // PyTypeObject *arg2 = Py_TYPE(obj)
        emit_load64_mem(Dst, arg2_idx, arg1_idx,  offsetof(PyObject, ob_type));
        //| type_version_check, arg2_idx, la->type_ver, >1
        emit_cmp32_mem_imm(Dst, arg2_idx, offsetof(PyTypeObject, tp_version_tag), (unsigned int)la->type_ver);
        emit_cmp64_mem_imm(Dst, arg2_idx, offsetof(PyTypeObject, tp_version_tag), (unsigned int)la->type_ver);
        dasm_put(Dst, 323);
```

pyston/pyston_lite/pyston_lite_autoload.pth
```diff
@@ -0,0 +1 @@
+import os, sys; exec("""sys.path.append("/home/kmod/pyston/pyston/pyston_lite"); ("PYSTON" in os.environ) and __import__("pyston_lite").enable()""")
```
undingen: hard coded user `kmod` in path. Would be nice if this file would get generated or you find another way to get the path.

pyston/pyston_lite/Makefile
```diff
@@ -51,6 +51,7 @@ BENCH:=../macrobenchmarks/benchmarks/bm_flaskblogging/run_benchmark.py --legacy
 # note: the testsuite doesn't pass for my Ubuntu system cpython, but it does pass
 # if I reconfigure this makefile to use a manually-built cpython
 testsuite: env
+	PYSTON=1 ./env/bin/python -c 'import test.support; test.support.check_impl_detail = lambda **kw: False; import test.test_code; test.test_code.test_main()'
```
undingen, line 54: It can't find test_code for me not sure how I have to generate it.

undingen (commented):
Cool new era starting I'm excited about the possibilities :sparkles:. I think it would also be easy to get `pyston_lite` working on macos (as long as we stick to gcc) later on.

LGTM
I agree with sharing as much code as possible and adding ifdefs looks like the best way to proceed.
Added a few smaller comments

undingen:
I just noticed some problems: I think we have to check `type->tp_flags &Py_TPFLAGS_VALID_VERSION_TAG` in every IC because stock
`PyType_Modified` does not set `tp_version_tag` to 0 :(.
Also there is a problem with it overflowing: 3.8 will just clear the internal `method_cache` and reassign the numbers which I think we can't really fix just have to hope 32bit is enough :(.

Update: looks like starting from 3.10 they also set `type->tp_version_tag = 0`

pyston/pyston_lite/Makefile
kmod: I had to do `sudo apt-get install libpython3.8-testsuite` since I think Ubuntu strips out the tests from their package. If you test this against a manually-built cpython I think it just works.

kmod:
Ok I pushed a bunch of new commits, would you mind taking a look?

pyston/pyston_lite/Makefile
undingen: a thanks that fixed it

undingen:
LGTM.

Python/aot_ceval_jit.c
```diff
@@ -915,6 +915,11 @@ static void switch_section(Jit* Dst, Section new_section) {
 // branches to false_branch on inequality else continues
 |.macro type_version_check, r_type_idx, type_ver, false_branch
 ||#ifdef PYSTON_LITE
+|| JIT_ASSERT(Py_TPFLAGS_VALID_VERSION_TAG == (1UL << 19), "need to update these offsets");
+@ARM|| emit_load32_mem(Dst, get_tmp_reg(r_type_idx), r_type_idx, offsetof(PyTypeObject, tp_flags))
+@ARM|  tbz get_tmp_reg(r_type_idx), 19, false_branch // Test Bit Zero: jump to false_branch if bit 19 is zero
```
undingen: `@ARM|  tbz Rw(get_tmp_reg(r_type_idx)), #19, false_branch`

Python/aot_ceval_jit.c
```diff
@@ -915,6 +915,11 @@ static void switch_section(Jit* Dst, Section new_section) {
 // branches to false_branch on inequality else continues
 |.macro type_version_check, r_type_idx, type_ver, false_branch
 ||#ifdef PYSTON_LITE
+|| JIT_ASSERT(Py_TPFLAGS_VALID_VERSION_TAG == (1UL << 19), "need to update these offsets");
+@ARM|| emit_load32_mem(Dst, get_tmp_reg(r_type_idx), r_type_idx, offsetof(PyTypeObject, tp_flags))
```
undingen: missing ;

Python/aot_ceval_jit.c
undingen: There are two places which reference `PyObject_IsTrueProfile` on arm64 I think it's safe to change them to `PyObject_IsTrue`

undingen (commented):
found a few small problems on arm

PR #218, opened 2022-05-10
pyston-lite safety fix

kmod:
this is one commit on top of #217 

we have to clear the existing opcaches when we take over the interpreter.

I am vaguely worried that there are additional problematic cases that I haven't thought about, but I think it's ok that there may still be some _PyEval_EvalFrameDefault frames on the C stack because they can only create new opcaches on function entry.


undingen (approved):
To be honest quite a hack but I also don't know a better way. 
If we store our opcache somewhere else in the `_PyCode_GetExtra` it would fix the problem but than we would also require more memory if we have the old and new code cache enabled at the same time. So I'm fine with your current solution and if it causes problems we can change it later on

kmod:
Oh I like that idea of using _PyCode_GetExtra, I think I'll switch to that

PR #220, opened 2022-05-13
jit: inplace modify temporary float objects

undingen:
This identifies owned floats with refcount==1 which can be inplace modified.
Works very well on math heavy code:
```
nbody                    1.25x faster
scimark_fft:             1.09x faster
scimark_sparse_mat_mult: 1.15x faster
spectral_norm:           1.08x faster
```
While Pyston is not focusing on on speedup math code this is quite a small and simple change
which shows quite large speedups on different workloads so I think it's worth adding it.

I had a earlier version which also added a path for longs but this had more mixed results because it only handled some of the longs which have been positive not using the small int cache etc... I will try adding a free list instead.

### DIFF
--- a/Include/internal/pycore_code.h
+++ b/Include/internal/pycore_code.h
@@ -140,10 +140,17 @@ typedef struct {
     PyTypeObject* type;  /* borrowed type */
 } _PyOpcache_Type;
 
+typedef struct {
+    PyTypeObject* type;  /* borrowed type */
+    unsigned char refcnt1_left; /* how many times had the left operand a refcnt of 1 */
+    unsigned char refcnt1_right;/* how many times had the left operand a refcnt of 1 */
+} _PyOpcache_TypeRefcnt;
+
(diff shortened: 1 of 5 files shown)


kmod:
lgtm!

will this work for the INPLACE case though -- won't those objects still have a reference from the locals array? or I guess they will get rejected for being borrowed?

The Faster CPython folks discussed possibly clearing the variable before doing the inplace operation: https://github.com/faster-cpython/ideas/discussions/162#discussioncomment-1962614

undingen:
I think this will catch cases like this where the right hand side is temporary:
```python
v1[0] -= dx * b2m
v1[1] -= dy * b2m
v1[2] -= dz * b2m
v2[0] += dx * b1m
v2[1] += dy * b1m
v2[2] += dz * b1m
for (r, [vx, vy, vz], m) in bodies:
    r[0] += dt * vx
    r[1] += dt * vy
    r[2] += dt * vz
```
It's not optimal, would be better when we directly modify the destination operand but should still save creating a temporary variable.

I will add support for the simple case `v1 -= dx` like we already have for unicode concat in a new PR.

PR #221, opened 2022-05-13
add free list for 'long' objects

undingen:
packports a change from 3.11 which splits up the cases for common single digit longs and creates a free list for them (the free list change is not part of 3.11).
1.02x mean speedup on pyperformance


kmod:
lgtm! that's a pretty nice improvement

PR #222, opened 2022-05-13
pyston-lite: put our opcache in co_extra

kmod:
Previously I was using the normal fields in PyCodeObject, but this
is risky because pyston-lite and CPython have conflicting ideas
of what those fields mean.

So instead, put the data in co_extra.

But only for pyston-lite; pyston-full still uses the default fields.


undingen (approved):
LGTM

PR #224, opened 2022-05-19
pyston_lite: type checks have to check for Py_TPFLAGS_VALID_VERSION_TAG

undingen:
CPython does not set `tp_version_tag` to 0 on modification so we have todo this additional check instead
fixes crash in `pyston/test/caches.py`


kmod:
thanks for catching this!

PR #225, opened 2022-05-19
Add CI test for pyston_lite

kmod:


undingen (approved):
great, also only running 7mins :)

PR #226, opened 2022-05-20
jit: fix arm64 call patching in pyston_lite

undingen:
pyston_lite has a bit different address space layout:
JIT code tries to be close to cpythons code. But our JIT helper functions which are located in
the pyston_lite.so are very far away.

E.g.:
```
0xfffff7810xe558 JIT_HELPER_LOAD_METHOD_CACHED
0x0000000001e5a924 is where the call instruction is located
```
This means we have to generate a mix of relative and absolute calls on arm too.
Maybe in the future it makes more sense to just use emit a jump table and emit indirect jumps on arm
but this fixes the issue for now.


kmod:
thanks for catching this!

PR #227, opened 2022-05-20
CI: add pyston_lite arm64 qemu bot

undingen:
Takes about 1h to run so will not slowdown our CI runs. 
I had to disable a few tests which segfault with qemu (which is not very encouraging) but I think we should try out and if it often causes issues disable it again. But it's super nice to get at least some feedback about arm without having to run the tests all the time manually.


kmod:
good thinking! I can take a look into why those tests are failling

PR #228, opened 2022-05-20
jit: inplace modify more cases

undingen:
Added a few missing cases and support for lists analog to the unicode support

- float:
  - handles now `a += b` and `a = a + b` where `a` has refcnt = 1 (in interpreters case it's actually 2 at time of the binop)
- unicode:
  - handled now `b = get_tmp_str() + "str"` where return value of `get_tmp_str()` is a owned str with only one reference
  - refactored code
  - better stats
- list:
  - added support for it

Stats for scons on godot:
```
jit: num inplace concat: 29 hits: 258066 misses: 7
```

our benchmarks and performance seems to be unchanged.

### DIFF
--- a/Python/aot_ceval_jit.c
+++ b/Python/aot_ceval_jit.c
@@ -2475,7 +2475,7 @@ static void emit_inline_cache_loadattr_entry(Jit* Dst, int opcode, int oparg, _P
 // special inplace modification code for float math functions
 // can modify either the left or right operand
 // returns 0 if generation succeeded
-static int emit_special_binary_op_inplace(Jit* Dst, int inst_idx, int opcode, int oparg, RefStatus ref_status_left, RefStatus ref_status_right, int load_store_left_idx) {
+static int emit_special_binary_op_inplace(Jit* Dst, int inst_idx, int opcode, int oparg, RefStatus ref_status_left, RefStatus ref_status_right, int load_store_left_idx, PyObject* const_right_val) {
     switch (opcode) {
         case BINARY_ADD:
         case BINARY_SUBTRACT:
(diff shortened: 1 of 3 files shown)


Python/aot_ceval_jit.c
```diff
@@ -2486,35 +2489,48 @@ static int emit_special_binary_op_refcnt1(Jit* Dst, int inst_idx, int opcode, in
...
+    // if profiling showed that the right operand can be inplace modifed more often use it instead
+    int load_store_left = load_store_left_idx != -1 && cache->refcnt2_left >= cache->refcnt1_right;
+    if (!cache->refcnt1_left && !cache->refcnt1_right && !load_store_left) {
         return -1;
+    }
+    // should we inplace modify the right operand?
+    int use_right = !load_store_left && ref_status_right == OWNED && cache->refcnt1_right >=  cache->refcnt1_left && cache->refcnt1_right >= opcache->optimized/2;
+    if (!use_right) { // try the left
+        // the inplace modified reference must be owned otherwise the refcnt==1 does not mean it's temporary.
+        if (ref_status_left != OWNED)
```
kmod, line 2513: For `a += b`, won't the lhs be a borrowed reference?

Python/aot_ceval_jit.c
```diff
@@ -2541,25 +2557,151 @@ static int emit_special_binary_op_refcnt1(Jit* Dst, int inst_idx, int opcode, in
...
+    if (!opcache || !opcache->optimized) {
+        return -1;
+    }
+    _PyOpcache_TypeRefcnt* cache = &opcache->u.t_refcnt;
+    if (cache->type != &PyUnicode_Type && cache->type != &PyList_Type) {
+        return -1;
+    }
+    if (cache->type == &PyUnicode_Type) {
+        // we always emit this IC even if the refcount is not 1 and it can't be modified inplace
+        // because
```
kmod: this looks leftover?

Python/aot_ceval_jit.c
undingen: Good catch! fixing this increased the number of hits quite a lot on some benchmarks

PR #229, opened 2022-05-20
A few more optimizations

kmod:
On flaskblogging:
- LTO + PGO: about 0.8% combined (mostly lto)
- Reenable caches for classes with `__getattr__` but not `__getattribute__`: 2.4%
- Copy in some function definitions so that they can get inlined: 0.8%

I did the lto+pgo work in the makefile, I'll convert that over to setup.py next


kmod:
This brings it to 17% faster on my dev machine and 4.5% faster on the benchmarking machine

kmod:
It's about 9% faster on an ec2 machine

undingen (approved):
nice :+1:

PR #231, opened 2022-05-23
allow disabling BOLT on x86

git-bruh:
Hi, this PR allows explicitly disabling BOLT (enabled by default) on x86. I'm trying to build pyston on a system without BOLT installed (Trying to get pyston to be built as a regular system package instead of vendoring stuff in submodules)


kmod:
awesome, thanks!

PR #233, opened 2022-05-24
Migrate pyston-lite build system to setup.py

kmod:
I was avoiding doing this because I was worried about the development workflow, but it doesn't seem that bad.


kmod:
I added bolt support as well; it seems to help about 0.2% which I guess is expected because we only spend about 1.5% of our time in the interpreter in this particular benchmark.

undingen (approved):
LGTM, thanks for changing over to using setup.py

PR #234, opened 2022-05-24
pyston_lite: enable LA_CACHE_VALUE_CACHE_DICT for the case where the attribute comes from the class

undingen:
This catches some of the cases like this:
```python
    class Foo:
      a = 1
    foo = Foo()
    foo.a #<--
```
Unfortunately the instance dict will often change so this cache will frequently miss (but this also enabled the
normal caching code to handle it so as long as the instance dict is stable for at least a few iterations it will help).
Can you please confirm that you are not seeing a perf regression - my numbers are changing a lot but I did see a small performance increase.

### DIFF
--- a/Python/aot_ceval.c
+++ b/Python/aot_ceval.c
@@ -3425,12 +3425,16 @@ main_loop:
 
             _PyOpcache *co_opcache;
             OPCACHE_CHECK();
-            if (USE_STORE_ATTR_CACHE && co_opcache && likely(v)) {
+            if (USE_STORE_ATTR_CACHE && co_opcache && co_opcache->optimized && likely(v)) {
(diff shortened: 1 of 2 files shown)


kmod:
My numbers are more variable than I thought -- the first time I measured this PR I got a 2.1% improvement, and the second time I got a 1.3% improvement.

Either way it's still a lot more than I expected -- I thought it was going to be quite rare to look up type attributes on singletons, which seems like the main time that this would hit.

kmod:
Third time I got a 0.1% improvement.... haha

undingen:
> Third time I got a 0.1% improvement.... haha

Reminds me of this xkcd:
image

:-P

kmod:
:+1:

PR #239, opened 2022-06-02
Enable split-dict caches for pyston-lite

kmod:
perf is a bit variable but it looks like this helps about 2-3%, mostly from the from-type cache type.

The main idea is that I set it to guard on the value of ma_keys, since if we make the keys object immortal then keys can only be added, not removed from it. For the positive-cache types (where we want to guard that the attribute exists) we can just guard on the keys pointer value; for the negative-cache type (where we want to guard that the attribute does not exist, so that we can continue to get it from the type) we also need to guard on dk_nentries to make sure that no new keys were added.

### DIFF
--- a/Include/internal/pycore_code.h
+++ b/Include/internal/pycore_code.h
@@ -35,9 +35,8 @@ enum _PyOpcache_LoadAttr_Types {
     // (only used if the more powerful LA_CACHE_IDX_SPLIT_DICT or LA_CACHE_VALUE_CACHE_SPLIT_DICT is not possible)
     LA_CACHE_VALUE_CACHE_DICT = 0,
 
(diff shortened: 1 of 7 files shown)


undingen:
Wow this is awesome, I'm seeing much better cache stats for pyston lite and large performance increases (while memory usage only slightly increases if at all) :+1: also stats for regular pyston are similar so I don't think any regression sneaked in.
I fixed the issue and all tests passes: https://github.com/undingen/pyston_v2/commit/41abc178e59746327482187496a2efb225f2d1bf https://github.com/undingen/pyston_v2/runs/6705674670?check_suite_focus=true

undingen (approved):
awesome

PR #240, opened 2022-06-02
A new LOAD_GLOBAL cache

kmod:
This is on top of #239

It uses our OFFSET approach, where we store the offset in
the hashtable and then check to make sure the same key
is still stored there. This lets it work for cases that
the globals are being reassigned.

This new cache type is about twice as fast as no-cache, and
and the directly-cache-the-value type is twice again as fast.

Faster CPython does this by guarding on the keys version, which
we could also do (would be faster than this new cache type, and
would also allow caching builtin lookups when globals change),
but I don't think there's an easy way to get that to work for pyston-lite so I just did this one for now


undingen (approved):
LGTM

kmod:
I looked at the perf stats for flaskblogging, and it looks like we were spending very little time doing LOAD_GLOBAL beforehand.

kmod:
so I don't think there's more to do here, and I think I'm done with changes before the release

PR #241, opened 2022-06-03
use LOAD_GLOBAL cache for some LOAD_NAMEs

undingen:
this allows us to cache stuff in a repl session and at module scope e.g.
```python
def f():
    return 42
for x in range(100000):
    f() # we now cache the "f" lookup here
```


kmod:
Do you think we need to worry at all about how this would work for class definitions (the other use of LOAD_NAME that I'm aware of)? Maybe we could add a check to not emit ICs for that case?

otherwise lgtm!

undingen:
From looking at the source I thought that `LOAD_GLOBAL` and `LOAD_NAME` is the same when `f_globals == f_locals` and (it's a dict) and will emit same errors etc. But I could be wrong if you find time could you please try out the class definition case.

kmod:
I think they're probably the same, but what I meant was that for the class-definition case, where `f_globals != f_locals`, we'll constantly be failing these checks. I think it's unlikely that people do a lot of work inside a class definition but seems like at least for the IC case we could avoid this without too much work?

undingen:
ah yes makes sense will look into that

undingen:
I pushed an improved version which should speedup in the JIT the case where `f_locals != f_globals`

kmod:
Won't this still be a slowdown vs before this pr when `f_locals != f_globals`? I was imagining that the jit could know that it's jitting a function where `f_locals` is probably not the same as `f_globals` and then it wouldn't emit the inline cache at all.

It's not a big deal though since I don't think people do performance-sensitive work in class definitions.

undingen:
I think this case should be handled:
- if `f_locals != f_globals` we will not fill the inline cache in the interpreter
- in the JIT we only emit the `LOAD_NAME` IC when `co_opcache->optimized`
- so the JIT will just always call into the `LOAD_NAME` helper which now does not do any caching

kmod:
ah ok makes sense!

PR #242, opened 2022-06-03
pyston_lite: create cibuildwheel buildbot

undingen:
- the workflow needs to be manually started: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
- enforces that the mac package is not available on 10.15 where pyston_lite currently crashes
- unfortunately we only have x86 CI machines so have to generate the remaining packages manually
- will only show up as workflow after it's merged
- works for linux amd64 too but I disabled it because otherwise we have two ways to build a package and `make package_jit` will also work on arm.
 
for macos arm64 I just manually run the steps in a terminal following the workflow file and created the wheel this way.

### DIFF
--- a/.github/workflows/build_pyston_lite_wheels.yml
+++ b/.github/workflows/build_pyston_lite_wheels.yml
@@ -1,0 +1,61 @@
+name: Build pyston_lite wheels
+
+on: [workflow_dispatch]
+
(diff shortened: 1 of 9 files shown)


kmod:
lgtm!

PR #244, opened 2022-06-03
Update version numbers 2.3.3->2.3.4

kmod:


undingen:
LGTM. I'm fine with calling it 2.3.4 or 2.4.

kmod:
I think calling it 2.4 might mean having to rebuild all of the conda packages, so I'd say let's call it 2.3.4

PR #245, opened 2022-06-03
Get bolt working with the manylinux wheel

kmod:
The bolt that has been built on the host does not typically work
inside the manylinux container. We could try to build bolt inside
the container, or maybe we could create a package and install it.

But it seemed easier to just generate a non-bolted wheel, then
run bolt on the host and repackage the wheel.


undingen (approved):
LGTM, thanks for fixing the arm issue. I also noticed it but you have been faster with fixing it :).
I'm not going to merge it in case you decide to name it 2.4 (but maybe because you already did the work we should just name it 2.3.4?)

PR #247, opened 2022-06-04
fix uninitialized var in LOAD_GLOBAL

undingen:
`_PyDict_GetItemByOffset(name)` was using a uninitialized var `name` in the interpreter - same code in JIT helper uses it initialized.

Stats stay the same on my quick test so the compiler may by chance have generated code which always initialized it.


undingen:
I just noticed that I also forgot to enable the LOAD_NAME cache for pyston_lite. 
With this two commits the stats for accessing a global at module scope are also looking good for pyston_lite.

kmod:
ah good catch!

after the release I'll try to combine the two opcache initialization functions so we don't have to remember to keep them in sync

PR #248, opened 2022-06-06
2.3.4: pump build_num for meta packages

undingen:
the use the python version number but reference the pyston version which means
we won't generate new meta packages which reference the new pyston version if the build_num stays the same.

I guess we could permanently fix this by integrating the full pyston build number into the build string
but this feels risky doing it now for the release without testing it more.

---
I already pushed the resulting packages `pyston`, `python` and `python_abi` to our pyston channel and it seemed to fix the installer issue where it would package the old version.

### DIFF
--- a/pyston/conda/python/meta.yaml
+++ b/pyston/conda/python/meta.yaml
@@ -1,4 +1,4 @@
-{% set build_num = 4 %} # caution: only reset when python_version changes
+{% set build_num = 5 %} # caution: only reset when python_version changes
 {% set pyston_version = "2.3.4" %}
 {% set python_version = "3.8.12" %}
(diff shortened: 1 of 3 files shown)


kmod:
looks good! It seems like the comment on those lines though is now the opposite of what we want?

undingen:
I added it because I think it's normally recommended to reset the build number to 0 when the version changes. But thats not possible for our recipes because they don't use the full version number everywhere so packages need to have a higher build number to get precedence. So I think we can only reset it when the python version and pyston version changes at the same time.

kmod:
oh I see, I read it as "only *change* when python_version changes"

PR #250, opened 2022-06-07
I guess rsync skips symlinks by default

kmod:
The build mostly worked except it skipped some of the files in the macrobenchmarks directory, so the bolt task would fail


undingen (approved):
thanks for fixing my `rsync` bug!

Issue #204, opened 2022-04-22
Pyston build for ubuntu 22.04


stuaxo:
Hi,
   Now Ubuntu 22.04 is out, is it possible to provide release a deb for it ?
(21.10 might be nice too)

Cheers
Stuart

undingen:
good idea, I will create 22.04 packages. We don't plan to create packages for non LTS versions right now so for 21.10 I recommend using the "portable release" archive.

stuaxo:
Cheers :)

PR #252, opened 2022-06-10
Build ubuntu 22.04 packages

undingen:
Unfortunately I run into some issues building 22.04 packages:
- our AOT PGO task is failing with - (producing empty profiles - I disabled it for now):
 >warning: ../aot_pic/aot_prof/default_16542604123857942053_0.profraw: failed to uncompress data (zlib)
error: no profile can be merged

- I'm seeing a lot of:
 >Warning -- files was modified by test_quopri
  Before: []
  After:  ['default.profraw']

I have not tested the performance of the resulting packages. But wanted to put it up as draft before my holiday in case 22.04 packages become a priority in the meantime.

I also noticed that one CFLAGS is different when using 22.04 and can be reverted by setting `DEB_BUILD_OPTIONS="reproducible=-fixfilepath"` but it did not fix the issues I encountered

We have a open issue for 22.04 packages at: https://github.com/pyston/pyston/issues/204

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -547,7 +547,7 @@ ifeq ($(RELEASE),16.04)
 	echo 10 > pyston/debian/compat
 	cd pyston; DEB_BUILD_OPTIONS=nocheck dpkg-buildpackage -b -d
 else
-	cd pyston; DEB_BUILD_OPTIONS=nocheck dpkg-buildpackage --build=binary --no-sign --jobs=auto -d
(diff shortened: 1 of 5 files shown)


kmod:
I've seen that second issue before and I think it's spurious -- the cpython testsuite checks to make sure that no files were created so it balks at the profile data files, which is maybe part of why the return code of the pgo task is ignored

PR #256, opened 2022-07-08
Install cargo into the github ci

kmod:
I'm not sure why this all of a sudden became required.
This is to build cryptography, a Rust extension, but we didn't
change the version of cryptography so I'm not sure why
the build process changed.

This also seems unrelated to the urllib3 flakiness that
I've been investigating.


kmod:
ok this looks ready for review

undingen (approved):
:+1:

PR #258, opened 2022-07-21
jit: fix bug in LOAD_NAME IC

undingen:
The jump referenced the previous defined label 4 and not the one following (direction was wrong).

Found this while looking for some other bug in the Python 3.7 support.


kmod:
:+1:

PR #260, opened 2022-07-27
Add call_function "aot trace" to pyston_lite

kmod:
I took another look at this optimization I tried to do last week and discovered the issue was that
I had only specialized function calls, not method calls (which are essentially the same but more common)

I think we could probably get another percent or two by adding specializations for a few other cases,
but I haven't done the investigation to know what they are.

---

Doesn't use any of the actual "aot" machinery; I just copied in
all of the relevant functions and specialized them by hand, and the
inliner takes care of merging them into a single function.

This commit only adds specializations for calling Python functions / methods.
It is not specialized for the argument count, unlike pyston-full.

About a 2% speedup on pyperformance and 1% on macrobenchmarks

### DIFF
--- a/Python/aot_ceval.c
+++ b/Python/aot_ceval.c
@@ -5224,7 +5224,7 @@ exit_eval_frame:
 // Entry point when executing a python function.
 // We check if we can use a JIT compiled version or have to use the Interpreter
 #ifdef PYSTON_LITE
-static PyObject* _Py_HOT_FUNCTION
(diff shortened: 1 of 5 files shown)


undingen (approved):
LGTM

undingen:
Looks like the merge commit also run into the problem which I (hopefully) fixed in #263 
https://github.com/pyston/pyston/runs/7650725655?check_suite_focus=true

PR #261, opened 2022-08-01
testsuite: fix unopt build by updating gevent

undingen:
the gevent version we used has a "Cython >= 3.0a6" requiriment but is incompatible with the new Cython == 3.0a8. Use a newer gevent to work around the problem.

```
  Error compiling Cython file:
  ------------------------------------------------------------
  ...
  cdef load_traceback
  cdef Waiter
  cdef wait
  cdef iwait
  cdef reraise
  cpdef GEVENT_CONFIG
        ^
  ------------------------------------------------------------

  src/gevent/_gevent_cgreenlet.pxd:181:6: Variables cannot be declared with 'cpdef'. Use 'cdef' instead.
```


kmod:
thanks for fixing this!

PR #262, opened 2022-08-01
pyston-lite: add Python 3.7 support 

undingen:
Finally after way to long time the 3.7 support seems to be ready, I did not do any performance comparison yet:

- had to copy in `ceval_gil.h`, `condvar.h` and `opcode_targets.h` from CPython 3.7 and added a 37 to the end of the file name
- created `aot_ceval_includes.h` which contains the huge number of includes the JIT and interpreter share
- biggest change is that in 3.7 the interpreter keeps a variable around which is called `why`. I added this now also to the JIT
which gets a pointer to it passed in from the interpreter, stores it on the stack and modifies it when required.
- moved `call_function_ceval_fast` from `lib.c` to `aot_ceval.c` because it can access definition from it more easily without having to copy more stuff.
- tested it on linux x86 and arm with cpython 3.7.13 successfully
but could not get it compiled on macos arm because of some I think pyston unrelated incompatibility with 3.7 (did not 100% confirm).
- fixed some compile warnings
- everything using vector calls is disabled because it seems not supported in 3.7
- added a linux pyston-lite 3.7 CI bot


Include/internal/aot_ceval_jit_helper.h
```diff
@@ -118,6 +118,13 @@ __attribute__((flatten)) JIT_HELPER_WITH_OPARG3(CALL_FUNCTION_EX_KWARGS, kwargs,
 JIT_HELPER_WITH_OPARG2(MAKE_FUNCTION, qualname, codeobj);
 JIT_HELPER_WITH_OPARG(FORMAT_VALUE);
 
+
+#if PY_MINOR_VERSION == 7
```
kmod: It's a really small nit, but if it's not too hard could we have these also guard on the major version? My impression is that `PY_VERSION_HEX` was made for doing these sorts of guards. Or you could just add a static assert that `PY_MAJOR_VERSION == 3` at the top of the file :P

Python/aot_ceval_jit.c
```diff
@@ -4679,6 +4833,9 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
 @ARM| sub sp, sp, #(num_callee_saved+num_stack_slots)*8
 @X86| sub rsp, num_stack_slots*8
 
+#if PY_MINOR_VERSION == 7
+    emit_store64_mem(Dst, arg4_idx /* =value */, sp_reg_idx, NUM_MANUAL_STACK_SLOTS * 8);
```
kmod: A tiny improvement would be to do something like `#define WHY_SLOT NUM_MANUAL_STACK_SLOTS` so it's more clear what this line is doing.

kmod:
This looks awesome! I'm not very familiar with the 3.7-vs-3.8 differences, but this pr looks very reasonable

kmod (approved):

Include/internal/aot_ceval_jit_helper.h
undingen: good point I replaced them with `#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION == 7` checks which I find easier to read then the `PY_VERSION_HEX` if I want to check for all 3.7 version

undingen:
Good suggestions! I have fixed them, and modified the code slightly so that 3.7 has now also a function call trace.
I'm waiting for the CI to finish before merging.

WoolenWang:
nice job~

PR #263, opened 2022-08-03
jit: fix possible access of freed memory

undingen:
During code generation we accessed a borrowed reference which
in some cases could be invalid. This was not a problem while executing the code because
it's correctly guarded etc but JIT code generation itself has the problem.

I could have fixed it by caching `descr->ob_type->tp_descr_get` when filling the cache entry
but unfortunately there is not enough space left in the cache entry so decided to just replace
it with a indirect call.

This got triggered sometime by the `unopt_build` CI bot in the django testsuite.


kmod:
Ah ok that makes sense, thanks for fixing this!

PR #264, opened 2022-08-09
Make sure these are absolute paths

kmod:
Bolt will silently fail to do anything meaningful if you provide a relative
path to the profiling output, since it will write all the profile data to
temporary directories that get deleted.

I ran into this when trying to port bolt to cpython, so I wanted to check
that we're using abs paths for pyston-lite. Looks like we are but I added
a call to os.path.abspath just to be sure


undingen:
Good catch! Not sure why macos suddenly failed but guess must be a fluke because it's not using BOLT.

kmod:
passed when rerun => no problem :stuck_out_tongue:

PR #266, opened 2022-08-16
add Python 3.9 support to pyston_lite

undingen:
Main changes:
- `COMPARE_OP` is split into 3 different opcodes `COMPARE_OP`, `CONTAINS_OP`, `IS_OP`
  - I made sure that `IS_OP` is using our optimized code
  - but the other function just branch to `PyObject_RichCompare`/`PySequence_Contains`
   because we can't do much more currently in pyston_lite because we don't have traces for them.
- most controversial: JIT generates the following code for the eval breaker checks:
  `emit_mov_imm(Dst, interrupt_idx, (unsigned long)&tstate->interp->ceval.tracing_possible);`
  - I think this means that we don't support subinterpreters because the address is hardcoded and not fetched everytime from the tstate.
   I thought this is not a problem for now because the feature to me seems WIP and if we want to support it I fear there are more cases where we would have to pass tstate and not fetch it via `PyThreadState_GET()`.
- cpython 3.9 is per default compiled `-fvisibility=hidden` which means we can only call public API.
  - This means I had to copy in more code into lib.c
  - the copied `_PyAsyncGenValueWrapperNew()` is not using a free list while cpython does
  - found another way to access `lookdict_split` which also works on 3.8 (but not on 3.7)

I did not check performance but changed our CI to run tests on 3.9


kmod (commented):
I think this is fine re subinterpreters -- I believe that objects can't be shared between subinterpreters, and since the jit code is specific to the code object it should be fine? Even if that's not the case I think "subinterpreter-incompatible for now" is perfectly fine.

Python/aot_ceval.c
```diff
@@ -5857,7 +6384,11 @@ _PyEval_EvalFrame_AOT_JIT(PyFrameObject *f, PyThreadState * const tstate, PyObje
     f->f_executing = 0;
     tstate->frame = f->f_back;
 
+#ifndef PYSTON_LITE
     return _Py_CheckFunctionResult(NULL, retval, "PyEval_EvalFrameEx");
+#else
+    return retval;
+#endif
```
kmod, line 6369: Just curious, why did this have to get added?

Python/aot_ceval.c
```diff
@@ -105,7 +105,11 @@ static PyObject * unicode_concatenate(PyThreadState *, PyObject *, PyObject *,
 /*static*/ PyObject * special_lookup(PyThreadState *, PyObject *, _Py_Identifier *);
```
kmod, line 105: I'm assuming that all the new code is this file is just a copy from cpython? The additions all seem reasonable to me, but if there were any changes made to the code then maybe we can call it out in some sort of way?

Python/aot_ceval_jit.c
```diff
@@ -2311,12 +2383,18 @@ static int emit_special_store_subscr(Jit* Dst, int inst_idx, int opcode, int opa
...
 // returns 0 if generation succeeded
 static int emit_special_compare_op(Jit* Dst, int oparg, RefStatus ref_status[2]) {
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION <= 8
     if (oparg != PyCmp_IS && oparg != PyCmp_IS_NOT) {
         return -1;
     }
+    int is_cmp = oparg == PyCmp_IS;
+#else
+    // for 3.9 we only call this function from IS_OP which means this but be a IS or IS_NOT comparison.
+    int is_cmp = oparg == 0;
```
kmod, line 2394: Sorry, does `oparg == 0` signify something?

Python/aot_ceval_jit.c
```diff
@@ -3998,24 +4085,22 @@ void* jit_func(PyCodeObject* co, PyThreadState* tstate) {
...
             RefStatus ref_status = deferred_vs_pop1(Dst, arg1_idx);
             deferred_vs_convert_reg_to_stack(Dst);
-            void* func = get_aot_func_addr(Dst, opcode, oparg, 0 /*= no op cache */);
+            void* func = NULL;
+#if PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 9
+            if (opcode == LIST_TO_TUPLE) {
+                func = PyList_AsTuple;
+            } else
+#endif
+            func = get_aot_func_addr(Dst, opcode, oparg, 0 /*= no op cache */);
```
kmod, line 4103: Would you mind adding braces around this line? Or something to make it a bit more clear that this is (sometimes) the body of the else block.

pyston/pyston_lite/setup.py
```diff
@@ -14,8 +14,8 @@
 NOPGO = "NOPGO" in os.environ
 BOLTFLAGS = "BOLTFLAGS" in os.environ
 
-if sys.version_info[:2] != (3, 7) and sys.version_info[:2] != (3, 8):
-    raise Exception("pyston-lite currently only targets Python 3.7 and 3.8")
+if sys.version_info[:2] != (3, 7) and sys.version_info[:2] != (3, 8) and sys.version_info[:2] != (3, 9):
```
kmod: So many versions! Now we could do `(3, 7) <= sys.version_info[:2] <= (3, 9)`

kmod (approved):
This looks awesome! I'm unfamiliar with these changes from 3.9 so I looked mostly at the jit, but everything looks reasonable.

Maybe at some point we could add our external testsuites to the pyston-lite ci run? The complexity is growing and my familiarity is decreasing so it might be nice to have more testing.

Python/aot_ceval.c
undingen: In 3.9 it got one more argument (tstate) and because we define it as a NOP I thought it's easier to just not call it instead of adding ifdefs for the cases with different number of arguments.

Python/aot_ceval_jit.c
undingen: Yes they use it for `IS_OP` and `CONTAINS_OP` to signify that it's a `is not` / `not in` comparison:
```
int res = (left == right)^oparg;
PyObject *b = res ? Py_True : Py_False;
```
I will add a comment to explain it.

Python/aot_ceval.c
undingen: Yes I tried to call them out if I really made changed but skipped some small things like removing "static"

undingen:
good ideas :+1: . 
Will add more tests when I finished 3.10

Python/aot_ceval_jit.c
kmod: Ohh I see -- the name `is_cmp`, especially with the alternate definition `oparg == PyCmp_IS`, read to me as "is this an `is` comparison", as opposed to dealing with negation

PR #267, opened 2022-08-26
Python 3.10: add Pyston_lite support

undingen:
main changes:
- oparg of jump targets and f_lasti is now counted in instructions not byte offset into the code object as before (means must args are now /2 because `sizeof(_Py_CODEUNIT) == 2`
- eval breaker checks are now done on entry, back edges and call opcodes.
- need to carry around a 'trace_info' object which the JIT keeps in a register similar to the `eval_breaker` it uses for other versions
- few more opcodes most are used for pattern matching
- free lists are now not stored in static global variables anymore but are part of the `tstate` which means I was able to make use of the free list in `_PyAsyncGenValueWrapperNew()` again


Python/aot_ceval.c
```diff
@@ -2508,7 +2933,34 @@ _PyEval_EvalFrame_AOT_Interpreter(PyFrameObject *f, int throwflag, PyThreadState
             } \
         } \
     } while (0)
-
+#else
+#define HANDLE_JUMP_BACKWARD_OSR() \
```
kmod: I think it's worth trying to separate the difference between these two versions into a separate conditionally-defined function/macro to avoid the duplication (it took me a while to find the difference)

kmod:
This looks great! I'm not sure this was less work than 3.7 but I'm glad it was less difficult :)

kmod (approved):

PR #268, opened 2022-08-30
pyston_lite: update make package

undingen:
- `make package` builds now wheels for all supported python versions
- autoload package is now a binary package


kmod:
looks great!

PR #269, opened 2022-09-01
pyston_lite: test all supported pyston_lite versions on the x86 macos CI 

undingen:
run cpython testsuite on macos for pyston_lite 3.7-3.10


kmod:
:+1:

PR #271, opened 2022-09-13
pyston_lite: run external testsuites on amd64

undingen:
- skip django and pandas on 3.7 because of dependency problems
- it's important that pyston_lite* get's installed into the virtualenv the test run else we will not use pyston lite. I verified that it works by checking in a test that `'pyston_lite' in sys.modules == True`


kmod:
awesome thanks so much for doing this!

PR #272, opened 2022-09-14
github CI: build all x86 linux and macos pyston_lite packages for a release

undingen:
After this action is merged we can go to "Action" -> "Build pyston_lite wheels" -> "run workflow" and trigger a build of all pyston_lite x86 packages. About 2h later we can download the wheels from the "Artifacts" page. This packages can be used for a release.

Because there is not arm runner the arm packages need to build manually.

### DIFF
--- a/.github/workflows/build_pyston_lite_wheels.yml
+++ b/.github/workflows/build_pyston_lite_wheels.yml
@@ -3,54 +3,44 @@ name: Build pyston_lite wheels
 on: [workflow_dispatch]
 
 jobs:
-  build_pyston_lite_wheels:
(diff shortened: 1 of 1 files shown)


undingen:
To confirm that it's working I did push this commit into my personal repo and run it and it did succeed:
https://github.com/undingen/pyston_v2/actions/runs/3055653807

I think you can't take a look at the artifacts it generated so I attached them (this is just a test for the release we want to update the package version etc...):
[artifact.zip](https://github.com/pyston/pyston/files/9570184/artifact.13.zip)

kmod:
very cool!

PR #273, opened 2022-09-21
make package: add suport for MacOS

undingen:
building packages outside the CI on Mac is annoying because it seems only one python official python version can be installed at a time which means the script will modify your system python... (only if `CI=1` is set else it will just abort when it can't find correct python version)

This changes make it very easy for me to build the ARM64 MacOS packages: I just need to run ''CI=1 make package" and it will automatically install the correct python versions and output the wheels


kmod:
:+1:

PR #276, opened 2022-09-26
Improve the auditwheel commands

kmod:
- auditwheel repair the new "pyston" (non-"lite") packages
- manually handle the autoload packages which it will error on due to not having any binary files


undingen (commented):
:+1:

Issue #279, opened 2022-10-12
Alpine/musllinux wheels


nikolaik:
Humbly requesting to start building musllinux wheels (if possible). When trying to install pyston under alpine, it fails since wheels are only built for manylinux.

I'm not too familiar with pyston to know if that is feasable or not.

Pypa has images to help build weels for musllinux, like being done here: https://github.com/qqaatw/aws-crt-python-musllinux/blob/main/.github/workflows/build.yml#L22 or one could probably start using [cibuildwheel](https://github.com/pypa/cibuildwheel).

PR #280, opened 2022-10-14
Add musllinux builds

nikolaik:
This is an attempt at solving #279 

Does the build step that builds bolt also need to run inside of of the musllinux container?

Note: I temporarily disabled building for all pythons to allow CI to run a bit faster. I'll enabled again after verifying the approach


kmod:
Thanks for taking a stab at this!

I don't see how to run the wheels-building action on your PR, but I think you could run it in your fork.

I think it's fine if bolt and the wheels are built differently.

PR #284, opened 2022-11-02
pyston3.8 symlinks

### LINKED ISSUE #283: pyston-full: symlinks for pyston3.8
It would be very useful to have a `pyston3.8` symlink in the docker image, debian installs, etc.  I suspect it is just an oversight.

I don't want to use `python3.8`, since that precludes coexistence with CPython 3.8.  And if pyston-full were ever backported to 3.9 or beyond, I'd want those variants also to coexist, so `pyston3` is not appealing.

belm0:
closes #283

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,12 @@ ARCH:=$(shell uname -m)
 -include Makefile.local
 
 .PHONY: all
-all: pyston3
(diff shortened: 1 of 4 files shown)


belm0:
test_sys_settrace and test_syslog failures seem to be unrelated, and occurring in other recent PR's

kmod:
Sorry for the delay, I was looking into the test failures but got pulled away.

Thanks for the change!

Issue #39, opened 2021-05-22
`platform.python_implementation()` returns "CPython"


rocky:
In Pyston 2.2: 

```
$ python
Python 3.8.8 (heads/rel2.2:6287d61, Apr 29 2021, 15:46:12)
[Pyston 2.2.0, GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
rocky's .pythonrc loaded
>>> import sys, platform
>>> platform.python_implementation()
'CPython'
>>> "Pyston" in sys.version
True
```

The code from v1 could be used though: https://github.com/pyston/pyston_v1/blob/6d392ba7a5882139e8a6ad8bf5165abd5f32f2d3/from_cpython/Lib/platform.py#L1465

kmod:
This is a rather delicate issue akin to picking a user-agent string, and I'm not 100% sure we can change this. We've run into code that is something like

```
if implementation == "CPython":
  doNormalThing()
elif implementation == "PyPy":
  doSpecialPyPyThing()
else:
  raise Exception("not supported")
```

though I don't remember where `implementation` in my example came from.

My takeaway is that we should have some tests so that we can know whether we can change something like this, because right now I think it's fairly risky.

rocky:
I wasn't suggesting that the code that I had in the Python interactive session is how would fix this (although it is the workaround I am using - the burden of my code outside is my worry). 

If you look at #40 the way this is handled is not to do any testing at all but return "Pyston" because if the code comes from this code base it is Pyston.

kmod:
What I'm saying is that while it's easy to code up this change (as you did in #40), it will require good testing to know if this is something that we want. The exact semantics of these functions are defined by how they are used by the community, and we've done a little bit of testing to find the best set of compatibility markers to get different projects to work. Unfortunately I don't remember the exact details at this point, but there is code like the pseudocode I listed earlier that could break with your change: your change will signal to certain code that we are not compatible with CPython. I wish there was a `sys.is_compatible_with_cpython` flag but since there's not this might be one of the ways that we have to signal compatibility.

mattip:
> The exact semantics of these functions are defined by how they are used by the community, and we've done a little bit of testing to find the best set of compatibility markers to get different projects to work

[...]

PR #286, opened 2022-11-19
Update the readme to talk about pyston-lite more

kmod:


README.md
```diff
@@ -36,16 +41,20 @@ These have not been heavily tested, so if you run into any issues please report
 
 ## Checking for Pyston at runtime
 
-Our current recommended way to see if your Python code is running on Pyston is to do `hasattr(sys, "pyston_version_info")`.
+To check for pyston-full, one can run `hasattr(sys, "pyston_version_info")`.
```
belm0, line 44: by the way, why not the standard property `sys.implementation.name`?


```python
>>> import sys
>>> sys.implementation.name
'pyston'
```

kmod: IMO switching on the implementation name is bad practice because it's not clear exactly what it means -- what if some day someone forks Pyston and calls it something else? Should all the "is this pyston" checks fail?

We ran into this in a different sort of way before: https://github.com/pyston/pyston/issues/39

